PR #71, opened 2020-02-28
add features

disregardfiat:


disregardfiat:
This adds a github API wrapper and initial user organization workflow automation that is commented out in the likely integration point

PR #155, opened 2020-05-12
Feature/add event

### LINKED ISSUE #92: Build "Add Event"  page
## Description
As an Admin user, I want to be able to create a new Event on the fly, regardless of event type (hack night, conference, summit, weekend meetup, etc.). I should be able to add a new event to the DB from the VRMS admin dashboard so that we can check-in users, track volunteer hours (custom to the event), and so on. 

A new event may or may not be a hacknight, and may include "sessions" which is why there is the below dependency.

## Dependency
#64 - add new fields to the "events" table (event_type, session)
Needs User Access Levels in system to be able to identify who can add an event

## Acceptance Criteria

shiningjustice:
Adds add event feature (resolves #92 ). 

Adds `AddNew`, `AddEvent` component. 

`AddEvent` allows an event to be added to either `events` table or `recurringEvents` table. The same data is passed in for either event (any recurring event is assumed to be repeated on a weekly basis). Fields/dropdowns are chosen/populated from Kelsey's [figma wireframe](https://www.figma.com/file/9c6pxjImN8MjHAZLH7zxYm/VRMS-Mocks?node-id=22881%3A2739).

Notes on form inputs: 
- selecting type `hacknight` 
  - displays radio buttons to select hacknight location. 
    - selecting `online` for hacknight location enables display of video conference link field 
  - disables `city` field (city is automatically selected based on hacknight location)
- selecting true for `remote` field removes city/state fields and replaces with video conference link 
- new event dates can easily be added using the plus/minus button by the dates (adds a date a week from the previous date). this was created before we had the recurring events table and can be removed if it doesn't add significance anymore.


shiningjustice:
@sarL3y 

Changes
- Updates projects field in object in POST body per updates to `event` and `recurringEvent` models 
- Fixes setVideoConferenceLink not working issue by adding an arrow function. Removes created handle function.
- Adds appropriate user-facing error message if project isn't selected and eventType = 'Hacknight' (previously would return ~ "Cannot get property '_id' for undefined");

PR #189, opened 2020-06-08
Feature/direct checkin returning user

### LINKED ISSUE #162: User Request: Remove extra step in Returning User check-in process
### Overview
As a Returning User, I want to enter my email address and check in with one action. The current iteration prompts the user to click a second button after their email address has been validated. The question displayed and the check-in button could be moved to the email validation page. 

### Action Items
- [x] Combine check-in functionality with email validation function, while accounting for the questions being asked 
- [x] Replace "Enter" button with "Check In"

shiningjustice:
Fixes #162 :

- Combines check-in functionality with email validation function, while accounting for the questions being asked
- Replaces "Enter" button with "Check In"

Changes: 
- `checkEmail` calls `submitReturning` in promises if user is found (`checkEmail` is only called for returning users)
- `submitReturning` optionally handles a passed in event, but requires the user to be passed in as an argument (previously it accessed user from the state). The `user` object is passed in directly because there is no direct way to asynchronously call `submitReturning` from `checkEmail` after `setUser` successfully updates the state.


shiningjustice:
@sarL3y newest commit address requested changes: 
- removes checkin question
- form and check in button persists until success page is loaded 

[Video demo of check-in for a returning user that has not answered the questions.](https://www.loom.com/share/5401c995ef7c4c2aa6f7e4dbcbf6c0dc)

PR #212, opened 2020-07-02
Fix unhandled promise warning

nickbeaird:
I am figuring out the development workflow here as I don't often use a forked repo myself. 

ISSUE:
After letting my development environment sit for a while, I was noticing the errors listed below in the logs start to fill my console. 

WORK DONE:
- I added some basic error handling to the functions throwing the error.

### DIFF
--- a/workers/closeCheckins.js
+++ b/workers/closeCheckins.js
@@ -56,9 +56,11 @@ module.exports = (cron, fetch) => {
     async function runTask() {
         console.log("Closing check-ins");
 
-        const eventsToClose = await sortAndFilterEvents();
-        // console.log(eventsToClose);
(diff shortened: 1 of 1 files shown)


nickbeaird:
I thought that adding something to use as an example to work through would be helpful for me learning. @sarL3y  you are welcome to review, or I will meet up with @clockwerkz here soon as well.

clockwerkz:
Cool! Alternatively you could wrap all those awaits in a try catch block, that way there's only one catch. Either way, thanks for adding in that error catch!

alex-anakin:
The `development` branch is fixed so the pull request should be merged into it. I edited the destination.

Issue #219, opened 2020-07-15
Fix broken UI on the admin dashboard


OlgaBilogurova:
### Overview

The admin dashboard longer than a background because of charts. Exist redundant horizontal & vertical scrollbars, scrolled background, margin just for top.

https://monosnap.com/file/zDsUkKEpTlY2KGfANRbEjD0zABOST8

OlgaBilogurova:
Progress: Done.

Before: 
[click](https://monosnap.com/file/kS4iVlDwXdZifL6Qsnr3GOec0Ca5US)

After:
[click](https://monosnap.com/file/onoi3UC4uh9oXYDAb2XmDCqwNy4XLA)
[click ](https://monosnap.com/file/qtm1dQ3n9dQ4yhwEhyNd43i1VNFckM)

Pull request: https://github.com/hackforla/VRMS/pull/220

PR #259, opened 2020-08-18
adds empty field check and disable btn styling

clockwerkz:
Fix for issue #219 : disabling button when the email text box is empty.

I added in a test to make sure the email string isn't empty. I also added some :disabled styling in Checkin.scss. 
Note: the component structure of the checkin/login forms can use a refactor.

disabled_button_sm

### DIFF
--- a/client/src/sass/CheckIn.scss
+++ b/client/src/sass/CheckIn.scss
@@ -134,11 +134,10 @@ input[type=radio] {
     height: 40px;
     margin-top: 20px;
     margin-bottom: 36px;
-    border: 3px solid black;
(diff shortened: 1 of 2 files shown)


client/src/sass/CheckIn.scss
```diff
@@ -147,6 +146,12 @@ input[type=radio] {
             background-color: black;
             color: white;
         }
+        &:disabled {
```
nickbeaird, line 149: Good to know. Cool trick and way to use css. https://css-tricks.com/the-sass-ampersand/. 🙇

Issue #262, opened 2020-08-20
Add a new event to the recurring events worker


alex-anakin:
### Overview
Add a new event to https://github.com/hackforla/VRMS/blob/master/workers/createRecurringEvents.js worker:
* type: HackNights
* day: Tuesday
* time: 6 pm - 8 pm

This a hotfix so a pull request should be created to `master` branch. The deadline is Monday, before 7 pm.

clockwerkz:
I wrote up a script that will enter in a recurring event into the database. I'm testing it with a test recurring event set for Friday (08/21/2020). If that works out, I'll mod the script for the new recurring event on Tuesdays. In the meantime, There's some additional info I could include when I create the event, such as:
name, hacknight, eventType, description, and if it's tied to any existing projects
This is optional info, but would definitely be helpful.
Something to note: Currently, the recurringEvent model's date property is not required; this *should* be as the backend pulls this date to check if it needs to generate an event from.

clockwerkz:
**Progress** -Pending Complete:  I went ahead and entered in the event into the recurringevent collections. Will close after the cron job runs correctly and creates an event for the meeting on Tues (08/25/2020). 
Also found a bug with the events generator, will submit a PR via the development branch.
**Blockers** - None
**Availability** - After 6pm M-F

Rankazze:
In reviewing/validating the Live HackforLA.org site  we need to fix our typo from "for"  to "or"  in the following: "On boarding for Happy Hour"
@alex-anakin @clockwerkz 
Thanks

Rankazze:
Confirmed the selection loaded in the pick list.

PR #265, opened 2020-08-22
Hot fix/entering recurrent event

clockwerkz:
This is in reference to issue #262

I noticed that the process of creating events from the recurringevents collection will crash if there isn't a location property in a recurringevent doc. Currently, the code tests for properties within the location, but if location itself isn't supplied, it will crash. I went ahead and broke out the location field into it's own conditional to first check if location actually exists on the recurringevents document first.

I also have a script ready to run and add the new event requested by @Rankazze for Tuesday 6:00-8:00PM Pacific.


clockwerkz:
I will be withdrawing this PR to the Master branch and resubmitting to the development branch.

Issue #273, opened 2020-08-28
Attempting to generate a new events from the recurring events collection will cause an exception error if a document does not have a location property.


clockwerkz:
### Overview
There is a cron job that periodically scans the recurringevent collections for events to occur on the current day. It then attempts to create new event entries into the events collection. If a given recurrentevent document does not have a location property, it crashes the operation with an exception error.

### Action Items
Include a test for the location property instead of assuming it always exists.

Also, in the short term, make sure all of the recurring events in the production database have a location property.

jbubar:
No progress. was not able to make the time to look over this, in the deep way it needs

peteplass:
@jbubar Please update.

maburke09:
@jbubar please add a role label to this issue

PR #274, opened 2020-08-28
fix for issue #273

clockwerkz:
PR fix for #273 - Server will crash if there is a recurringevent without a location field.


workers/createRecurringEvents.js
```diff
@@ -90,6 +85,13 @@ module.exports = (cron, fetch) => {
                         startTime: filteredEvents[i].startTime && newEventDate,
                         endTime: filteredEvents[i].endTime && newEndTime,
                         hours: filteredEvents[i].hours && filteredEvents[i].hours
+                    } 
+                    if (filteredEvents.hasOwnProperty("location")) {
+                        eventToCreate.location = {
+                            city: filteredEvents[i].location.city && filteredEvents[i].location.city,
+                            state: filteredEvents[i].location.state && filteredEvents[i].location.state,
+                            country: filteredEvents[i].location.country && filteredEvents[i].location.country
+                        };
```
nickbeaird, line 94: I see that we are using a for loop using a counter. This seems like a good temporary fix. Have you thought of converting this loop to a for of loop instead? 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

clockwerkz: Yep, that's a good suggestion..would help with readability. :)

clockwerkz:
Well, it was a good thing you pointed that out because I failed to add [i] to filteredEvents.location!

I went ahead and changed the loop to a for of loop. It's a good idea to try and leave the code base a little neater than how you found it. :)

workers/createRecurringEvents.js
nickbeaird: I have a quick questions for my own learning. Does each field have to have the double value of VALUE && VALUE. Maybe I am misunderstanding, but why is that needed?

nickbeaird: After doing a little research, I see that the value on the right is used when I try adding two string objects in the shell (see my screenshot). For readability, can we make the values that are repeated the single value

i.e. 
hacknight: filteredEvent.hacknight
vs 
hacknight: filteredEvent.hacknight && filteredEvent.hacknight,

Then you may need to review the date values listed there and try and figure out which one you want to list. It might be a turnary value of one that exists OR something else. 

image

clockwerkz: This is a good point - I believe the idea is following the pattern to test the property first to see if it exists before using it's value. But in this case, it defeats the purpose because whether the result is true or false, the property is still being created..

This can definitely be simplified, but probably something we should break out and address as a code refactor at some later point..the sole purpose of this fix is to address that there is a double reach into a property that is causing a exception error if no location is provided at all.

clockwerkz: If we were to use this pattern correctly, I think the idea would be something like  if (obj.hasOwnProperty("someProp") { newObj.someProp = obj.someProp; }
This would only add the property of someProp with it's value if it exists. 
The current method of checking during the assignment will result in a value like "false" or undefined for that property. 
Thinking about it some more, maybe this **should** be addressed with this fix.. but I would want to have a discussion on the data model as it is and it's lack of "required" properties first.

nickbeaird: That makes sense. Okay yeah, I think that this link might be helpful in looking at validation of schema. https://stackoverflow.com/a/38616988/5900471

For the time being, if this has the desired outcome, then I will approve this PR. I think that Alex will want to review and merge as well.

clockwerkz: Sounds good - yeah I would prefer PRs to add fixes or features should only address one item at a time, I feel. I'm good with whatever Alex decides on this. :)

PR #287, opened 2020-09-04
SPIKE: Magic link

nickbeaird:
This Spike is to do some exploratory research into adding a Magic Link login functionality to the application. 

High Level:
1. User can sign up for the application at /api/auth/signup. 
  -> There is validation and cleaning of the payload
  -> There is verification of email and basic data
  -> An email is sent to  the user's email for verify the email.
  -> There are tests
2. Admin users can log into the application. Non-admin users cannot. 
  -> same as above
3. We have a test email account. Permissions can be provided upon request. 
4. Refactor the tests to help make working easier. 
5. Move logic from routes to middleware and controllers
6. Improve the import module system for the DB.


alex-anakin (approved):
Great work! I tested the branch locally, it works as described. Using nodemailer is exactly what we need to continue with Amazon SES.

PR #289, opened 2020-09-11
Refactor/user profile

clockwerkz:
NOT INTENDED TO MERGE:

I'm exploring ways we can use Auth and Hooks to manage state globally across the app. We can utilize Higher Order Components to wrap any component that needs to pull state from the state globally. 
I've wrapped the App in it's own HOC to establish both the Auth and User Contexts, and in the UserProfile Component I'm using the Context.Consumer to bring in the data and methods from the user Context.

### DIFF
--- a/backend/app.js
+++ b/backend/app.js
@@ -23,8 +23,9 @@ app.use(morgan("dev"));
 // Hide sensitive Header data
 // app.use(helmet());
 
-// Cross-Origin-Resource-Sharing
-// app.use(cors());
+//Cross-Origin-Resource-Sharing
+const cors = require('cors');
+app.use(cors());
 
 // WORKERS
 const runOpenCheckinWorker = require("./workers/openCheckins")(cron, fetch);
--- a/client/src/pages/UserProfile.js
+++ b/client/src/pages/UserProfile.js
@@ -35,7 +35,7 @@ export default (function (WrappedComponent) {
     return (props) => {
         return (
         <UserContext.Consumer>
-            {(context) => (<WrappedComponent {...context}/>)}
+            {(context) => (<WrappedComponent {...props} {...context}/>)}
         </UserContext.Consumer>
         )
     }
--- a/client/src/pages/UserProfile.js
+++ b/client/src/pages/UserProfile.js
@@ -5,7 +5,7 @@ import UserEvents from '../components/presentational/profile/UserEvents';
 import UserTeams from '../components/presentational/profile/UserTeams';
 import { UserProvider, UserContext } from '../context/userContext';
 
-const Profile = (props) => {
+const UserProfile = (props) => {
     const { user, setUser, removeOption, events, teams } = props.context;
     useEffect(() => {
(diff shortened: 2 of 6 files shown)


backend/app.js
```diff
@@ -23,8 +23,9 @@ app.use(morgan("dev"));
 // Hide sensitive Header data
 // app.use(helmet());
 
-// Cross-Origin-Resource-Sharing
-// app.use(cors());
+//Cross-Origin-Resource-Sharing
+const cors = require('cors');
+app.use(cors());
```
nickbeaird, line 28: I am wary of CORS, but likely from a lack of information. 

So from what I am seeing online, people love or hate this addon. We also had it in the application earlier, but it was commented out before I started. We have a CORS issue in the app on localhost only it seems. We have the backend and frontend on differing ports for localhost development, but then we change things in production. So adding in this on Production MIGHT cause a vulnerability. That is what I thought before though. 

To combat the localhost issues, we originally had "poxy": "htttp://localhost:4000" in the package.json. I moved that out to a .env file variable REACT_APP_PROXY and then added that info to setupProxy.js. I am not sure if adding a proxy is a good thing or bad thing, but I did it. The implementation that I had at least allowed us to change ports on the application, which I think is important. So these somewhat conflict. 

I keep seeing a lot of debate on whether or not to use CORS in relation to cookies. I started adding the token to a cookie for the authentication stuff. I also see that from the tutorial's that I have been using, that I have been setting "Access-Control-Allow-Origin:". Basically, I really want to better understand our API protection strategy and fix this. a but more. Here is a stack overflow article that talks about it https://stackoverflow.com/a/9726921 (I don't fully understand). And here is a tutorial that I used somewhat and that I want to use further to make our application more secure, https://levelup.gitconnected.com/how-to-implement-csrf-tokens-in-express-f867c9e95af0

I would greatly appreciate your opinions, and this may be an offline conversation.

client/src/pages/UserProfile.js
```diff
@@ -5,37 +5,39 @@ import UserEvents from '../components/presentational/profile/UserEvents';
...
 
-const UserProfile = (props) => {
-
+const UserProfile = ({ user, setUser, removeOption, events, teams }) => {
     useEffect(() => {
-        // fetchData();
+        fetch('http://localhost:6999/api/users/5e1d2346316d2f00172ef066')
+            .then(res => res.json())
+            .then(data => setUser(data))
+            .catch(err => console.log("Error fetching:", err));
```
nickbeaird, line 13: Clean and I love it. I am a fan of async await, but this is nice.

nickbeaird (approved):

backend/app.js
clockwerkz: I agree - I re-instated it as a temporary measure and I should have highlighted that. The CORS enable is not meant to be part of the PR. We are going to need to look at how we can safely handle requests across multiple domains as different sites may make use of our open API endpoints (like the website, as an example). We might just be able to whitelist them somehow..

clockwerkz:
Not intended to commit to development

PR #290, opened 2020-09-11
Magic link 2

### LINKED ISSUE #267: Get rid of Firebase
Details will be added later

https://reallifeprogramming.com/how-to-implement-magic-link-authentication-using-jwt-in-node-8193196bcd78
https://medium.com/one-more-thing-studio/how-to-make-magic-links-with-node-1d164c036e29

nickbeaird:
STEPS TO VERIFY: 
1. Pull my repo.
2. docker-compose build & docker-compose up
3. Log into the localhost:3000 page for the main app
4. Select the login button. Provide your email.
5. Log into your email and use the link provided. The link should direct you to the login page.
6. If you are not redirected to the login admin screen, then just refresh or login to the main page and press login again.

TODOs:
- Alex to determine if the approach is valid.
- Determine if we need to add a logout button now.
- Possibly make the user experience better with a redirect.
- Review the Tinyurl redirection for localhost
- Whatever else you think.


alex-anakin (changes requested):
1. I tried to log in under my own email. Firstly I got `User does not exist`. Then I made a small fix to the mongoose `find` command and got `Invalid permissions`. And here I'm confused because Mongoose returns `[Object object]` instead of actual user value. I used `util.inspect` so it is not an output issue.
2. Login for non-existing user email displays "You don't have the correct access level" but should "Please enter the correct email address".

nickbeaird:
I updated the branch to work with the most recent conflicts. 

1. Pull my repo.
2. docker-compose build & docker-compose up
3. Log into the localhost:5000 page for the main app
4. Select the login button. Provide your email.
5. Log into your email and use the link provided. The link should direct you to the login page.
6. If you are not redirected to the login admin screen, then just refresh or login to the main page and press login again.

TODOs:

- Alex to determine if the approach is valid.
- Determine if we need to add a logout button now.
- Possibly make the user experience better with a redirect.
- Review the Tinyurl redirection for localhost
- Whatever else you think.

alex-anakin (approved):

PR #303, opened 2020-09-16
Frontend linter 2

### LINKED ISSUE #293: Fix errors and warnings from compiler
The application has a number of compiler errors as well as runtime errors. At a high level, a great deal of these errors look to be resolved easily with enforcing eslint standards. 

Frontend issues: When you start the react app, a number of warnings and compile errors fill the console. 

Backend issues: When running the application, a number of uncaught async calls breaks the application. 

Overall: Code formatting and linting tools will help us code review faster, and make sure that we are focusing on the feature and not the syntax. Additionally, the tools can help give feedback faster in a programatic way, rather than relying on our own opinions.

nickbeaird:
This adds code formatting and linter rules. The linter and code formatter are not enforced at this time, but can be used manually if anyone wants to use these tools on their branch. 

Work Done:
I am removing refactor. 
Eslint configured in npm/yarn (not enforced)
Pretter configured in npm/yarn (not enforced)
Prettier configured and disabled on VSCode editor settings.
Prettified and eslint out-of-the-box changes for the admin page only. 

Possible next steps:
Olga to add admin page unit tests. 
Add tests and refactor. 
Review and possibly enforce linters and prettier. 

How to Test:
1. Open the PR locally
2. cd client
3. yarn run eslint FILENAME) --fix
4. yarn prettier --write FILENAME
5. Run in VSCod
4. You will see a number of errors in the console (see screenshot below)

### DIFF
--- a/client/src/pages/AdminDashboard.js
+++ b/client/src/pages/AdminDashboard.js
@@ -1,17 +1,17 @@
-import React, { useState, useEffect } from 'react';
-import { Redirect } from 'react-router-dom';
+import React, { useState, useEffect } from "react";
+import { Redirect } from "react-router-dom";
 
-import useAuth from '../hooks/useAuth';
+import useAuth from "../hooks/useAuth";
 
-import '../sass/Dashboard.scss';
-import UpcomingEvent from '../components/presentational/upcomingEvent';
-import EventOverview from '../components/presentational/eventOverview';
-import DonutChartContainer from '../components/presentational/donutChartContainer';
-import Loading from '../components/presentational/donutChartLoading';
+import "../sass/Dashboard.scss";
+import UpcomingEvent from "../components/presentational/upcomingEvent";
+import EventOverview from "../components/presentational/eventOverview";
+import DonutChartContainer from "../components/presentational/donutChartContainer";
+import Loading from "../components/presentational/donutChartLoading";
 
 const AdminDashboard = (props) => {
   const auth = useAuth();
-  const defaultChartType = 'All Events';
+  const defaultChartType = "All Events";
   let uniqueEventTypes = new Set();
   let hackNightUniqueLocations = new Set();
 
@@ -24,31 +24,14 @@ const AdminDashboard = (props) => {
   const [chartTypes, setChartTypes] = useState(null);
 
   // Volunteers SignedIn By Event Type
-  const [
-    totalVolunteersByEventType,
-    setVolunteersSignedInByEventType,
-  ] = useState({});
-  const [
-    totalVolunteerHoursByEventType,
-    setVolunteeredHoursByEventType,
-  ] = useState({});
-  const [totalVolunteerAvgHoursByEventType, setAvgHoursByEventType] = useState(
-    {}
-  );
+  const [totalVolunteersByEventType, setVolunteersSignedInByEventType] = useState({});
+  const [totalVolunteerHoursByEventType, setVolunteeredHoursByEventType] = useState({});
+  const [totalVolunteerAvgHoursByEventType, setAvgHoursByEventType] = useState({});
 
   // Volunteers SignedIn By Hacknight Property
-  const [
-    totalVolunteersByHacknightProp,
-    setVolunteersSignedInByHacknightProp,
-  ] = useState({});
-  const [
-    totalVolunteerHoursByHacknightProp,
-    setVolunteeredHoursByHacknightProp,
-  ] = useState({});
-  const [
-    totalVolunteerAvgHoursByHacknightProp,
-    setAvgHoursByHacknightProp,
-  ] = useState({});
+  const [totalVolunteersByHacknightProp, setVolunteersSignedInByHacknightProp] = useState({});
+  const [totalVolunteerHoursByHacknightProp, setVolunteeredHoursByHacknightProp] = useState({});
+  const [totalVolunteerAvgHoursByHacknightProp, setAvgHoursByHacknightProp] = useState({});
 
   // Volunteers To Chart
   const [totalVolunteers, setVolunteersToChart] = useState({});
@@ -60,9 +43,9 @@ const AdminDashboard = (props) => {
   async function getAndSetData() {
     try {
       setIsLoading(true);
-      const checkIns = await fetch('/api/checkins');
+      const checkIns = await fetch("/api/checkins");
       const checkInsJson = await checkIns.json();
-      const events = await fetch('/api/events');
+      const events = await fetch("/api/events");
       const eventsJson = await events.json();
 
       processData(eventsJson, checkInsJson);
@@ -73,7 +56,7 @@ const AdminDashboard = (props) => {
     }
   }
 
-  function processData(allEvents, allCheckIns) {
+  function processData(allEvents, allCheckIns){
     let processedEvents = processEvents(allEvents);
     let usersByEvent = collectUsersByEvent(allCheckIns);
     prepareDataForCharts(processedEvents, usersByEvent);
@@ -82,14 +65,14 @@ const AdminDashboard = (props) => {
   function processEvents(allEvents) {
     let events = new Map();
 
-    for (let event of allEvents) {
+    for (let event of allEvents){
       // Process legacy data with undefined 'hours' property because initially an event length was 3 hours
-      if (!event.hours) {
+      if(!event.hours){
         event.hours = 3;
       }
 
       // Define unique event types and process events without 'eventType' property
-      if (event.eventType) {
+      if(event.eventType){
         processEventTypes(event, 'eventType', uniqueEventTypes);
       } else {
         // Find events without 'eventType' property (30 events) and assign it
@@ -97,7 +80,7 @@ const AdminDashboard = (props) => {
       }
 
       // Extract events with 'hacknight' property & find unique locations in it
-      if (event.hacknight) {
+      if(event.hacknight){
         processEventTypes(event, 'hacknight', hackNightUniqueLocations);
       }
       events.set(event._id, event);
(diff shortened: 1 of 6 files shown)


nickbeaird:
Here is what the updated output for the Admin page component looks like

frontend_linter_warning_2

client/src/pages/AdminDashboard.js
```diff
@@ -149,65 +191,85 @@ const AdminDashboard = (props) => {
...
-      if(propName === 'eventType'){
+    uniqueTypes.forEach((el) => (result[el] = parseInt('0')));
+    for (const eventId of users.keys()) {
+      if (propName === 'eventType') {
         type = events.get(eventId).eventType;
         result[type] = users.get(eventId).length + result[type];
       }
 
-      if(propName === 'hacknight' && typeof events.get(eventId).hacknight !== 'undefined'){
+      if (
```
alex-anakin: lines less than 100 symbols shouldn't be formatted

client/src/pages/AdminDashboard.js
```diff
@@ -149,65 +191,85 @@ const AdminDashboard = (props) => {
...
-    let result = {};
-    uniqueTypes.forEach(el => result[el] = parseInt('0'));
-
-    for(let eventType of uniqueTypes.keys()){
-      let hours = totalVolunteerHours[eventType];
-      let volunteers = totalVolunteers[eventType];
-      let averageHours = (hours / volunteers);
-      if(!Number.isInteger(averageHours)) averageHours = +averageHours.toFixed(2);
-      volunteers > 0 ? result[eventType] = averageHours : result[eventType] = 0;
+  function findAverageVolunteerHours(
```
alex-anakin: lines less than 100 symbols shouldn't be formatted

client/src/pages/AdminDashboard.js
```diff
@@ -24,14 +24,31 @@ const AdminDashboard = (props) => {
   const [chartTypes, setChartTypes] = useState(null);
 
   // Volunteers SignedIn By Event Type
-  const [totalVolunteersByEventType, setVolunteersSignedInByEventType] = useState({});
-  const [totalVolunteerHoursByEventType, setVolunteeredHoursByEventType] = useState({});
-  const [totalVolunteerAvgHoursByEventType, setAvgHoursByEventType] = useState({});
+  const [
```
alex-anakin: Here and below lines that less than 100 symbols shouldn't be formatted. It's not clear why this statement should be divided into multiple lines.

client/src/pages/AdminDashboard.js
```diff
@@ -149,65 +191,85 @@ const AdminDashboard = (props) => {
...
+    uniqueTypes
+  ) {
+    const result = {};
+    uniqueTypes.forEach((el) => (result[el] = parseInt('0')));
+
+    for (const eventType of uniqueTypes.keys()) {
+      const hours = totalVolunteerHours[eventType];
+      const volunteers = totalVolunteers[eventType];
+      let averageHours = hours / volunteers;
+      if (!Number.isInteger(averageHours))
```
alex-anakin: let's use brackets for such `if` statements

alex-anakin (changes requested):

client/src/pages/AdminDashboard.js
nickbeaird: I may be misunderstanding what you are meaning, but I am not seeing that as a valid syntax request. Please let me know if I am missing something. I attached a screenshot to show what I am seeing. 

In addition, I have an opinion that using brackets will misconstrue the meaning. It would make the user think that the evaluation statement is an array, IMHO. 

image

client/src/pages/AdminDashboard.js
nickbeaird: First, I did a little research and I am not sure on how to set that request specifically. I did want to look at the config settings, but do you have a config setting that you are looking for? https://prettier.io/docs/en/options.html

Secondly, I feel good about these settings, but that may just be me having used formatters for a while now. There is an analogous project in Python now used widely called Black. For some reason, I had assumed that you were already using Prettier in your JS projects, so I did not try and sell it. That said, the Prettier and Black teams do a better job than I can on that. Please take a gander on the video and let me know your thoughts. https://www.youtube.com/watch?v=0Q4kUNx85_4

client/src/pages/AdminDashboard.js
nickbeaird: (see above related response)

client/src/pages/AdminDashboard.js
nickbeaird: (see above related response)

client/src/pages/AdminDashboard.js
alex-anakin: yeah, the video says: *If it fits in one line, stay in one line*
https://youtu.be/0Q4kUNx85_4?t=411

Also, we can try Google eslint config: https://github.com/google/eslint-config-google that uses Google JS style https://google.github.io/styleguide/jsguide.html

client/src/pages/AdminDashboard.js
alex-anakin: I made a mistake in my comment: braces, not brackets. Sorry about that. Here is an example:
```
if (condition) {
  statement
}
```
even if the statement is 1 line

nickbeaird: I think that this is a case of actually not looking too far into it. I am looking at the code and I think that it's broken. Here is the `development` branch with the same method and I am not sure what the intended goal of the lines at this time is. Do you have any thoughts on what these lines are intending to do @alex-anakin ? https://github.com/hackforla/VRMS/blob/development/client/src/pages/AdminDashboard.js#L190

In summary, I don't think that this is the formatter's issue, as much as code that is breaking prettier.

client/src/pages/AdminDashboard.js
nickbeaird: Yeah, I think that this is part of the Google V. Prettier Auto-formatter issue. It may not exactly relate to ESLint, but it is related. That same video around 8:40 that states to break on the first argument as part of Rule #2, which has a lot of nuance. There is separate logic and verification on lines 211 and 212, which makes me also agree that we should break them up and put them on separate lines. I also see this example relates pretty well to the part of the video found around 17:15, which states the operating logic is actually on lines 204, and the logic 205-216 is supporting that single truthy statement there. Organizing the code this way shows these lines as a block. Lastly, I can see this relating to the section of the video around 25:00 or so, which talks about bringing prettier onto an existing project. We happen to have formatted the Admin logic, which has a lot of buy in from developers. This I think causes a little more push back than usual, as it feels precious. Even though I state that, I would still highly suggest that we use these settings.

All of this said above, here are the take-aways that I think are super helpful. 
- The code keeps the line on one line for a single method call. It breaks on multiple method calls, or on multiple verification steps. It would not break if there was no && there. 
- The formatting is set to support the validation pathway. This is part of the second on Rule #2 as presented by the video and supports the code block path on 204.
- Formatting can feel frustrating at first. Everyone has their own preferences and automatically formatting a file can feel frustrating as it is a why. I truly think that this goes away with a little time. Part of that is getting used to the formatter and linter. The other part is that we all start coding a little more to standard when we use the tools from the start. That is another reason to actually implement these on all files. We have however disabled them for now. 
- The higher majority of projects these days are using Prettier. I chose AirBnB and Prettier as it is the most widely adopted settings for React projects such as ours. I think that these settings are pretty good. 
- As far as Google's setting for a project, there is one reason that I did not go the path of setting ESlint to Google. It was due AirBnB's high adoption of React, and Google's overall trends to support Angular. Configuring React with their linter settings required extra work.

client/src/pages/AdminDashboard.js
alex-anakin: I don't see anything broken in the code. The line that you mentioned is a function declaration.
```
function findAverageVolunteerHours(totalVolunteers, totalVolunteerHours, uniqueTypes) {
  // some calculations for component business logic
  return result;
}
```
What is wrong here for you?

I think you misunderstood my first comment. I want prettier to format this statement
`if (!Number.isInteger(averageHours)) averageHours = +averageHours.toFixed(2);`
to be the following:
```
if (!Number.isInteger(averageHours)) {
  averageHours = +averageHours.toFixed(2);
}
```

Regarding your comment about code that breaks the prettier I need to say that working code comes first formatting comes second.

client/src/pages/AdminDashboard.js
alex-anakin: Most important, we don't need any frustration here at all. This is not a job, everyone is a volunteer. So, no pushing.

The second idea is avoiding overhead. The rule **If it fits in one line, stay in one line** is totally clear for every developer. Going deeper is overengineering. For me first of all. 
Please, don't use multiple lines for function arguments, value assignments, and `if` conditions if it less 100 symbols long.

client/src/pages/AdminDashboard.js
nickbeaird: Yeah, sorry, I was referencing that line there,  `if (!Number.isInteger(averageHours)) averageHours = +averageHours.toFixed(2);`. That does not look to be proper syntax and looks to be breaking the autoformatter. My apologies for sending you the function that it housed in and not the line itself. Below is why I think that the syntax doesn't make sense, but I also think that this is the reason it is breaking the formatter. 

image

I would still defer to you on this. Are you expecting that the value is reassigned to it's positive? Or do you think it has another intention?

OlgaBilogurova: `if (!Number.isInteger(averageHours)) averageHours = +averageHours.toFixed(2);` 

There I determine that `averageHours` is not an integer. If true, convert it to a number of 2 digits after the decimal point. Method toFixed() returns a string representing a number in fixed-point notation. So, I converted it to a number using '+', because in total it will return NaN.

```
/**
 * Returns true if the value passed is an integer, false otherwise.
 * @param number A numeric value.
 */
isInteger(number: unknown): boolean;

/**
 * Returns a string representing a number in fixed-point notation.
 * @param fractionDigits Number of digits after the decimal point. Must be in the range 0 - 20, inclusive.
 */
toFixed(fractionDigits?: number): string;

 /**
  * Returns a supplied numeric expression rounded to the nearest integer.
  * @param x The value to be rounded to the nearest integer.
  */
 round(x: number): number;

```
If it's breaking the linter. The above line could be replaced with this one without any problem:

```
if (!Number.isInteger(averageHours)) {
  averageHours = Math.round(100 * averageHours) / 100;
}
```

nickbeaird: Yeah, this would work. Thanks @OlgaBilogurova, yeah this syntax is breaking the auto-formatter, so thanks for an alternative.

nickbeaird: Hey @OlgaBilogurova, does this look right? 
image

client/src/pages/AdminDashboard.js
nickbeaird: Yeah @alex-anakin, yeah I think the same thing stands that I am not seeing how to meet the goals that you are aiming for with these tools. If you can offer an alternative configuration for the tooling, then I am up for discussing that further. Otherwise, I feel like I have provided more than ample research and evidence on the how and why I chose these tools and configurations. 

For now, I will remove any of the auto-formatting on this file, and any VSCode configurations. This is truly something that I feel passionate about and bring a lot of expertise in, but it's not worth arguing over.

alex-anakin: Applying the formatting on the file was a great example of what prettier can do. My concern is formatting lines with a length of fewer than 100 symbols. I'm pretty sure there is a rule in prettier settings that controls this action and can be set to false.
Let's merge settings and then try to test eslinter and prettier.

alex-anakin (approved):

PR #307, opened 2020-09-22
Slack 

Jonathanko52:
Automatically sends zoom link before a scheduled event in the database. Requires someone on the Slack Administration to set up the app in the Slack API and an updated .env file.


backend/routers/slack.router.js
```diff
@@ -24,85 +20,53 @@ if (process.env.NODE_ENV !== "test") {
 }
 
 //Finds Id number of channel
-router.get("/findId", (req, res) => {
-  publishMessage();
-  findEvent();
-  // findProject();
-});
+router.get("/scheduleMessage", findCurrentEvents, scheduleMessage);
```
nickbeaird: I am a little confused by this. Are you aiming to GET or POST on this route?

backend/routers/slack.router.js
```diff
@@ -24,85 +20,53 @@ if (process.env.NODE_ENV !== "test") {
...
 
-//uses Id number to send message to said channel
-router.post("/postMeeting/:id", (req, res) => {
-  publishMessage1();
-});
+async function scheduleMessage(req, res, next) {
```
nickbeaird: I really like that you are creating a separate function for scheduling a message, but do we need to hardwire function to the api res object? I feel like we can either make this function not be Express Middleware and be only a function that we send the Events to, or maybe review if we can turn this into an Express endpoint and not be Middleware.

backend/routers/slack.router.js
```diff
@@ -24,85 +20,53 @@ if (process.env.NODE_ENV !== "test") {
...
-
-    console.log(result);
-  } catch (error) {
-    console.error(error);
-  }
-}
-
-async function findEvent(req, res) {
-  Event.find({})
+function findCurrentEvents(req, res, next) {
```
nickbeaird: I think that this relates to my statement above. I think that we are using a lot of space writing to the response object. We can likely use this logic inside the other call without.

backend/routers/slack.router.js
```diff
@@ -24,85 +20,53 @@ if (process.env.NODE_ENV !== "test") {
...
+  Event.find()
```
nickbeaird: I am not sure on what the performance of this query is, but we can do this in one call versus multiple. We can send a query to the DB to only return values where the date is greater than the date you want. The DB will then return a smaller query. We would not need to iterate over the longer array.

backend/workers/slackbot.js
```diff
@@ -1,93 +1,29 @@
...
-            .catch(err => {
-                console.log(err);
-            });
-        
-        console.log('Done sending');
+module.exports = (fetch, cron) => {
+  const url =
+    process.env.NODE_ENV === "prod"
+      ? "https://www.vrms.io"
+      : "http://localhost:4000";
```
nickbeaird: Please move any hard coded values to the .env file.

client/package-lock.json
```diff
@@ -0,0 +1,18703 @@
+{
```
nickbeaird: We are using yarn-lock files now rather than package.lock.json. Please fix this. 

1. Delete your node_modules
2. Delete your package.json files
3. Yarn install on all the folders 
4. Commit your yarn-lock file.

client/src/pages/Home.js
```diff
@@ -6,91 +6,109 @@ import ls from "local-storage";
...
+  const handleEventChange = (e) => {
+    setEvent(e.currentTarget.value);
 
-    useEffect(() => {
-        fetchEvents();
+    // setIsQuestionAnswered(true);
+  };
+  const slackButton = () => {
+    console.log("SLACK BUTTON");
+    fetch("api/slack/event?checkInReady=true");
```
nickbeaird: Do we need to hardcode checkInReady on all API calls?

client/src/pages/ProjectLeaderDashboard.js
```diff
@@ -62,7 +62,7 @@ const ProjectLeaderDashboard = () => {
           if (!item.userId.name) {
             item.userId.name = {};
             item.userId.name.firstName = item.userId.firstName;
-            item.userId.name.lastName = item.userId.lastName; 
+            item.userId.name.lastName = item.userId.lastName;
```
nickbeaird: I would review the function that contains this change and see if we can not mutate outside data, but rather create copies.

routers/events.router.js
```diff
@@ -0,0 +1,84 @@
+const express = require("express");
+const router = express.Router();
+
+const { Event } = require("../models/event.model");
+
+// GET /api/events/
+router.get("/", (req, res) => {
+  const { query } = req;
+
+  Event.find(query.checkInReady === "true" ? query : undefined)
```
nickbeaird: I don't feel like this is a protected endpoint. We are allowing the end user to put in any query if they have entered checkInReady === true. I think that we can add some better validation logic on this using middleware, rather than checking on this value here and passing in anything on the DB call if not.

routers/events.router.js
```diff
@@ -0,0 +1,84 @@
...
+    })
+    .catch((err) => {
+      console.log(err);
+      res.sendStatus(500).json({
+        message: `/GET Internal server error: ${err}`,
+      });
+    });
+});
+
+router.post("/", (req, res) => {
```
nickbeaird: Our GET query above has more validation than our POST.

backend/app.js
```diff
@@ -29,8 +29,11 @@ app.use(morgan("dev"));
 // WORKERS
 const runOpenCheckinWorker = require("./workers/openCheckins")(cron, fetch);
 const runCloseCheckinWorker = require("./workers/closeCheckins")(cron, fetch);
-const runCreateRecurringEventsWorker = require("./workers/createRecurringEvents")(cron, fetch);
-// const runSlackBot = require("./workers/slackbot")(fetch);
+const runCreateRecurringEventsWorker = require("./workers/createRecurringEvents")(
+  cron,
+  fetch
+);
```
alex-anakin: @Jonathanko52 Keep it on one line, please.

backend/routers/slack.router.js
alex-anakin: @Jonathanko52 In general, GET is for requesting any data, POST is for creating something. Please use POST for the endpoint because its purpose looks like "creating a new schedule for the message".
@nickbeaird The code review best practices recommend providing reason what caused the comment in case of opening a discussion. Or strict recommendation in case if you understand what is wrong.

backend/routers/slack.router.js
alex-anakin: @nickbeaird Looks like the PR author follows the current development style of the application. Don't forget about the existing legacy code. If you'd like to refactor this please create a followup task that can be done as part of resolving the application technical debt.
@Jonathanko52 skip this comment.

backend/routers/slack.router.js
alex-anakin: @nickbeaird Please take a look at my response to your previous comment.
@Jonathanko52 skip this comment.

backend/routers/slack.router.js
alex-anakin: @nickbeaird There are no multiple calls here. There is still the one call but it queries all events from the DB.
@Jonathanko52 In other words please add a condition for a date into the `find` method.

backend/workers/slackbot.js
alex-anakin: @Jonathanko52 I agree with Nick here.

client/src/pages/Home.js
alex-anakin: @nickbeaird This is a good question but it's better to address it to the legacy codebase. Please create a task for fixing this API endpoint and further client cleanup with a clear description of what should be done.
@Jonathanko52 skip this comment.

client/src/pages/AdminDashboard.js
```diff
@@ -24,14 +24,31 @@ const AdminDashboard = (props) => {
   const [chartTypes, setChartTypes] = useState(null);
 
   // Volunteers SignedIn By Event Type
-  const [totalVolunteersByEventType, setVolunteersSignedInByEventType] = useState({});
-  const [totalVolunteerHoursByEventType, setVolunteeredHoursByEventType] = useState({});
-  const [totalVolunteerAvgHoursByEventType, setAvgHoursByEventType] = useState({});
+  const [
```
alex-anakin: @Jonathanko52 This file should not be changed in the scope of the task. Please revert all changes for `client/src/pages/AdminDashboard.js` file.

client/src/pages/ProjectLeaderDashboard.js
alex-anakin: @nickbeaird The data from DB contains some legacy records that should be processed before further use. Original outside data is useless because it breaks rendering.
@Jonathanko52 skip this comment.

routers/events.router.js
alex-anakin: @nickbeaird Protecting endpoints is a good question but it's better to address it to the legacy codebase. Please create a task for fixing this API issue and further client cleanup with a clear description of what should be done.
@Jonathanko52 skip this comment.

routers/events.router.js
alex-anakin: @nickbeaird Of course. See my previous comment.
@Jonathanko52 skip this comment.

alex-anakin (changes requested):
@Jonathanko52 Please resolve my comments.

backend/routers/slack.router.js
nickbeaird: @alex-anakin, this function has been rewritten from an enpoint (req, res) and made into middleware (req, res, next). That makes this call now multiple calls, as it will be passed onto the next call. https://expressjs.com/en/guide/using-middleware.html.

backend/routers/slack.router.js
nickbeaird: For both @alex-anakin  and @Jonathanko52 , I asked the question, so that I could ask some follow up questions. I feel like you have both a goal of having a GET route with the findCurrentEvents method, as well as a post with scheduleMessage. I think that both can be true.

backend/routers/slack.router.js
nickbeaird: @alex-anakin, I respectfully must disagree. The use of Express middleware was only recently added by me, so there has not been any prior use of middleware to stipulate this as a current development style

From the Express docs, middleware is often used for validation of endpoints (https://expressjs.com/en/guide/using-middleware.html). I will restate that this method seems to be a really great method that could be used by a variety of endpoints. I don't think that using middleware is the most sustainable coding practice moving forward, and that pattern of mutating the response object in middleware endpoints can really make for challenging debugging later. I am suggesting that we use this method as either an endpoint (i.e (req, res)), or as a helper method.

backend/routers/slack.router.js
nickbeaird: @alex-anakin, I must again respectfully disagree. 

This comment that I made relates to the above statement in regards to the POST middleware, but is slightly different. The method here is findCurrentEvents, and stands out to me as a GET endpoint. I would assume that we are asking to return all events that are current and that could be an updated GET. This method however is using Express middleware with the use of the next variable in the method (see link below). The docs for Express Middleware suggest using Middleware for when you need to chain methods together across API endpoints, such as setting headers across a wide array of endpoints, or adding in verification for routes. This allows you to isolate your verification of endpoints away from the actual logic. I would again state that this method and the logic in it are best suited to be pulled together in the endpoint, and not used as Express middleware across files. 

https://expressjs.com/en/guide/using-middleware.html

client/src/pages/Home.js
nickbeaird: @alex-anakin, this is a new method, so it is not legacy code. I defer to you on this one though. I can only again suggest that we try and improve the code that we are adding, rather than adding it to future work.

Jonathanko52:
Changes have been addressed. Waiting for review.

PR #335, opened 2020-10-14
Update readme

OlgaBilogurova:


nickbeaird (approved):
Thanks and merging as it is so small.

PR #339, opened 2020-10-17
Hotfix/admin login

nickbeaird:
Issues: Developers have had intermittent issues of not being able to get their development environments running. This is often due to not having environment variables set. This updates the backend to error out stating which env variables are missing if any.  

Changes:
1. Fixes any issues where developers forget to add an environments variable files. Adds in a verification that all env variables are present.
2. Refactors the email controller to better communicate intent, and setup for local testing. 

How to test
1. Load the application including the backend. 
2. Remove a variable in the backend .env file
3. Load the application again and see the error that an env variable is missing. 

Caveats:
1. We cannot do this for the frontend as the assets are compiled before starting. This causing errors that are not easy to understand.

### DIFF
--- a/backend/app.js
+++ b/backend/app.js
@@ -9,7 +9,28 @@ const morgan = require("morgan");
 const path = require("path");
 const cookieParser = require("cookie-parser");
 
+// Import environment variables
 require("dotenv").config();
+
+// Verify environment variables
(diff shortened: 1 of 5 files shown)


alex-anakin (approved):
LGTM

PR #343, opened 2020-10-19
Redirect user to /admin after logging in

### LINKED ISSUE #342: Redirect user after successful login to /admin page
### Overview
Currently, only users with the `admin` access level can log in. Redirect the user after successful login to the `/admin` page.

alex-anakin:
Fixes #342.

### DIFF
--- a/client/src/pages/HandleAuth.js
+++ b/client/src/pages/HandleAuth.js
@@ -36,7 +36,7 @@ const HandleAuth = (props) => {
   } else {
     return (
       <div className="flex-container">
-        <div>Sorry, magic link is not valid</div>
+        <div>Sorry, the link is not valid.</div>
       </div>
     );
   }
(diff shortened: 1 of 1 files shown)


client/src/pages/HandleAuth.js
```diff
@@ -31,16 +32,16 @@ const HandleAuth = (props) => {
 
   let text;
   if (isMagicLinkValid == true) {
-    text = <p>Magic link is valid</p>;
+    return <Redirect to="/admin" />
   } else {
-    text = <p>Magic link is NOT valid</p>;
+    return (
+      <div className="flex-container">
+        <div>Sorry, magic link is not valid</div>
```
nickbeaird: If you are going for longer term, you might want to write "Sorry that link is not valid.". I am assuming that not every user will know what a magic link is. Either way, I am approving this PR. Feel free to merge as you see fit.

nickbeaird (approved):

client/src/pages/HandleAuth.js
alex-anakin: Good catch, thanks!

PR #352, opened 2020-10-21
346 use vrms host address for magic link

### LINKED ISSUE #346: Use the VRMS host address for building Magic Link
### Overview
We need to implement a mechanism for building Magic Link depending on the host of the application.

### Instructions
The easiest way to implement that is by using environment variables. Let's say if there is a variable `APP_HOST` in the `.env` file then the Magic Link mechanism uses its value. If there is no such variable then the default value `localhost:3000` should be used.

If there is a way that can resolve the VRMS host address automatically and it's not hard to implement then let's use it.

awhig:
Pass the request origin to the send email function which is used to generate the magic link

reqOrigin looks like. "http://localhost:3000"

Contains protocol, and port #, which is good.


backend/controllers/email.controller.js
```diff
@@ -53,9 +53,9 @@ const createProdSMTPTransport = async () => {
   return smtpTransport;
 };
 
-async function sendMail(smtpTransport, email, token) {
+async function sendMail(smtpTransport, email, token, origin) {
   const encodedToken = encodeURIComponent(token);
-  const emailLink = `https://tinyurl.com/nyqxd/handleauth?token=${encodedToken}&signIn=true`;
+  const emailLink = `${origin}/handleauth?token=${encodedToken}&signIn=true`;
```
nickbeaird, line 58: Did you test this using an email provider? The reason this was changed was that this broke for localhost. I was unfortunately glued to doing some sort of redirection.

nickbeaird: I am going to ask for you to at least verify that it works end2end for local development before approving.

nickbeaird (changes requested):

backend/controllers/email.controller.js
awhig: @nickbeaird Yes. This worked on my localhost.Magic link that was generated for me is below. Was sent to yahoo email and received. Linked worked for me on login.

http://localhost:3000/handleauth?token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6IjVmN2NmYzQ3ZjQ3ZGY2YzFmZmYwNTZlZiIsInJvbGUiOiJhZG1pbiIsImlhdCI6MTYwMzIzNjMxNiwiZXhwIjoxNjAzMjM3MjE2fQ.TzqekHoKPtHzkhDa2r8s8Y5OkGlPUVOfTQ-TlkUwvE8&signIn=true

alex-anakin (approved):
LGTM

nickbeaird (approved):

PR #361, opened 2020-10-24
Refactor backend part 1

### LINKED ISSUE #355: Refactor backend for magic link additions
The recent magic link work was done quickly and requires some follow up. We aim to increase the testing and confidence of the backend and make sure that the methods are well named and communicated. I am making sure that the endpoints work as expected and cleaning up the organization along the way.

### LINKED ISSUE #357: Make API HTTP codes and messages consistent across the application
### Overview
Currently, there is some inconsistency in HTTP codes, error messages, and success responses for different routers:
https://github.com/hackforla/VRMS/tree/development/backend/routers

### Action Items

- [ ] Use HTTP status codes mentioned in the description for fixing all responses.
- [ ] Make error handling consistent.
- [ ] Make error messages consistent.
- [ ] Clean up mistakes.

### Instructions
Success:
200 OK - Response to a successful GET, PUT, PATCH or DELETE.
201 Created - Response to a POST that results in a creation. Should return created object.

Errors:
400 Bad Request – This means that client-side input fails validation.
401 Unauthorized – This means the user isn’t not authorized to access a resource. It usually returns when the user isn’t authenticated.
403 Forbidden – This means the user is authenticated, but it’s not allowed to access a resource.
404 Not Found – This indicates that the API endpoint is not found.

502 Bad Gateway – This indicates an invalid response from an upstream server.
503 Service Unavailable – This indicates that something unexpected happened on the server side (It can be anything like server overload, some parts of the system failed, database failed, etc.).

nickbeaird:
Overview:
The goal of this PR was to refactor the backend to have standard controllers and improve the testability of the functionality. The changes here relate to tickets #357  and #355. The events, projects, and user routes have been refactored. This simplifies our router files to more easily communicate the standards of what our routes are doing. Additionally, moving the router logic to controller files allows us to mock out controller logic for unit and integration tests. 

Additional changes:
1. Remove the answers model and routes. They are not called anywhere. 
2. Remove any router errors that return internal service stack traces. This is a security standard and better practice. 
3. Add CRUD testing for any routes refactored. 

How to review:
1. Pull the PR
2. Run the integration tests with npm run tests in the backend. Run a specific test file if you have questions.
3. Review the controllers. 
4. Review the routers (Event, Email, Project, User). If you need more context, review the comments added on the files changed. 
5. Start the application. Review that an admin can login. Review that a user can navigate to the api routes. 

Fixes the issues in https://github.com/hackforla/VRMS/pull/340
Resolves https://github.com/hackforla/VRMS/issues/357, or at least for the most part. 
Resolves https://github.com/hackforla/VRMS/issues/355


backend/app.js
```diff
@@ -52,15 +51,14 @@ app.use(morgan("dev"));
 // app.use(cors());
 
 // WORKERS
-const runOpenCheckinWorker = require("./workers/openCheckins")(cron, fetch);
-const runCloseCheckinWorker = require("./workers/closeCheckins")(cron, fetch);
-const runCreateRecurringEventsWorker = require("./workers/createRecurringEvents")(cron, fetch);
+const runOpenCheckinWorker = require('./workers/openCheckins')(cron, fetch);
```
nickbeaird, line 53: I had to disable and re-enable the workers to get the integration tests working. I re-enabled them once I had test mocks in place and the remainder of the tests work.

backend/app.js
```diff
@@ -71,18 +69,17 @@ const projectTeamMembersRouter = require("./routers/projectTeamMembers.router");
 const slackRouter = require("./routers/slack.router");
 const authRouter = require("./routers/auth.router");
 
+app.use('/api/auth', authRouter);
 app.use("/api/events", eventsRouter);
-app.use("/api/checkins", checkInsRouter);
-app.use("/api/answers", answersRouter);
```
nickbeaird, line 76: No one is using the answers routers. I removed all answers references. The data in the questions is being held on the questions.

backend/controllers/email.controller.js
```diff
@@ -53,41 +62,61 @@ const createProdSMTPTransport = async () => {
...
+  const subject = 'Login to VRMS!';
+  const htmlMessage = `<a href=${encodedUri}>Login to VRMS</a>`;
+  sendEmail(email, subject, '', htmlMessage);
+}
 
-const emailController = {
-  sendUserEmailSigninLink,
+const EmailController = {
+  sendLoginLink,
+  sendEmail,
```
nickbeaird, line 121: Allows for two public email controller methods. The sendLoginLink sends a user an email for them to login. The sendEmail method it a utility method to send an email from our service. This second method may be worthy of being refactored into a utilities file, but this is where it stands for now.

backend/controllers/event.controller.js
```diff
@@ -0,0 +1,66 @@
...
+    const event = await Event.findByIdAndUpdate(EventId, req.body);
+    return res.status(200).send(event);
+  } catch (err) {
+    return res.sendStatus(400);
+  }
+};
+
+EventController.event_member_list = function (req, res) {
+  return res.send('NOT IMPLEMENTED: Event list UPDATE');
+};
```
nickbeaird: I am adding a method to implement later. I am advocating for a list of event members. These are likely the same, but may not be project members.

backend/controllers/event.controller.js
```diff
@@ -0,0 +1,66 @@
+const { Event } = require('../models');
+
+const EventController = {};
```
nickbeaird, line 3: Move all of the router controller logic to a controller. Also, rewrite into async/await for readability.

backend/controllers/index.js
```diff
@@ -0,0 +1,11 @@
+const EmailController = require('./email.controller');
+const EventController = require('./event.controller');
+const UserController = require('./user.controller');
+const ProjectController = require('./project.controller');
+
+module.exports = {
+  EmailController,
+  EventController,
+  UserController,
+  ProjectController,
```
nickbeaird, line 10: Rename controllers to match Pascal case convention used on Models. Event => EventController.

backend/controllers/project.controller.js
```diff
@@ -0,0 +1,67 @@
+const { Project } = require('../models');
+
+const ProjectController = {};
+
```
nickbeaird, line 4: Same as the EventController, send all of the router logic to a Controller. Use async/await. Also add tests.

backend/controllers/project.controller.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+};
+
+ProjectController.event_list = function (req, res) {
+  return res.sendStatus('NOT IMPLEMENTED: Get next Project for Project GET');
+};
+
+ProjectController.member_list = function (req, res) {
+  return res.sendStatus('NOT IMPLEMENTED: Get next Project for Project GET');
+};
```
nickbeaird: Adding for later, but looking to add these methods on the Project. 
* The member list of the Project is not the same as the event member list. People who attend an event are not explicitly VRMS members. 
* Also, we should be associating event members to the project. We are currently using a separate Model.

backend/controllers/user.controller.js
```diff
@@ -1,27 +1,115 @@
...
   // expires after half and hour (1800 seconds = 30 minutes)
-  return jwt.sign({ id: user.id, role: user.accessLevel }, CONFIG.SECRET, {
-    expiresIn: `${CONFIG.TOKEN_EXPIRATION_SEC}s`,
+  return jwt.sign({ id: user.id, role: user.accessLevel }, CONFIG_AUTH.SECRET, {
+    expiresIn: `${CONFIG_AUTH.TOKEN_EXPIRATION_SEC}s`,
   });
 }
 
-function createUser(req, res) {
+UserController.createUser = function (req, res) {
```
nickbeaird, line 103: TODO: We will need to figure out how to create a user with differing levels of access. At the moment, we have multiple routes.

backend/jest.setup.js
```diff
@@ -1 +1,8 @@
 jest.setTimeout(30000)
+
+// TODO: Refactor worker routes. These are setup to run cron jobs every time the app
+// is instantiated. These break any integration tests.
```
nickbeaird, line 4: As stated, we should revisit this pattern in future work.

backend/controllers/email.controller.js
```diff
@@ -53,41 +62,61 @@ const createProdSMTPTransport = async () => {
...
+const getEmailTransport = () => {
   let smtpTransport;
   if (process.env.NODE_ENV === 'development') {
-    smtpTransport = await createDockerSMTPSTransport();
-  } else {
-    smtpTransport = await createProdSMTPTransport();
+    smtpTransport = createMailhogSMTPTransport();
   }
+  smtpTransport = createGmailSMTPTransport();
+  return smtpTransport;
```
alex-anakin: function ALWAYS returns `gmailSmtpTransport`

backend/controllers/email.controller.js
```diff
@@ -37,8 +39,15 @@ const createDockerSMTPSTransport = async () => {
...
-const createProdSMTPTransport = async () => {
-  const accessToken = await emailCientToken();
+/** Sets up emails to be sent through Gmail.
+ *
+ * This allows us to use a Gmail account as our outgoing email. You can log
+ * into this email account to see that the sent messages includes your email.
+ *
+ * This requires setting up the Gmail account for use. https://tinyurl.com/y4farjwt
+ */
+const createGmailSMTPTransport = () => {
```
alex-anakin: for better readability, let's use `createGmailSmtpTransport` instead of `createGmailSMTPTransport`

backend/controllers/email.controller.js
```diff
@@ -20,12 +19,15 @@ const emailCientToken = async () => {
...
   return accessToken;
 };
 
-const createDockerSMTPSTransport = async () => {
-  // Send mail to Mailhog Docker container
+/** Sets up emails to be sent to the Mailhog docker container.
+ *
+ * The Mailhog container must be available for this to succeed.
+ */
+const createMailhogSMTPTransport = () => {
```
alex-anakin: for better readability, let's use `createMailhogSmtpTransport` instead of `createMailhogSMTPTransport`

backend/controllers/email.controller.js
alex-anakin: Sounds good, I like that we have the ability to send emails to users for instance as notifications.

backend/controllers/event.controller.js
```diff
@@ -0,0 +1,66 @@
...
+
+EventController.event_list = async function (req, res) {
+  const { query } = req;
+
+  try {
+    const events = await Event.find(query);
+    return res.status(200).send(events);
+  } catch (err) {
+    return res.sendStatus(400);
```
awhig, line 12: I think enums should be used for the return codes. It will provide readable text for each code, which would help with preventing coding errors. This package contains all the codes, maybe use this...  https://www.npmjs.com/package/http-status-codes

awhig (changes requested):
My only request is to use enums for the server status return values... See https://www.npmjs.com/package/http-status-codes

backend/controllers/email.controller.test.js
```diff
@@ -1,6 +1,8 @@
-const emailController = require('./email.controller');
+const EmailController = require('./email.controller');
```
alex-anakin, line 1: let's add a simple test for getEmailTransport that it returns correct transport depending on the environment

backend/controllers/event.controller.js
alex-anakin: Good idea, but let's do this in a separate PR

backend/controllers/event.controller.js
alex-anakin: As discussed in meetings, we will not have an event members list here. Please remove it.

backend/controllers/project.controller.js
alex-anakin: As discussed, we don't need it. Please remove.

backend/controllers/user.controller.js
alex-anakin: yeah, it was discussed

backend/routers/auth.router.js
```diff
@@ -12,18 +13,18 @@ router.use(function (req, res, next) {
...
+  '/signin',
+  [authAPIValidator.validateSigninUserAPICall, verifyUser.isAdminByEmail],
+  UserController.signin,
 );
 
-router.post("/verify-signin", userController.verifySignIn);
+router.post('/verify-signin', UserController.verifySignIn);
 
-router.post('/me', [authJwt.verifyCookie], userController.verifyMe);
+router.post('/me', [verifyAuth.verifyCookie], UserController.verifyMe);
```
alex-anakin: From my side, `verifyAuth` is not a very appropriate name. It starts from a verb and sounds like a method name. What about `AuthVerifier` or `AuthUtil`? Or chose what you like in a noun style.

backend/routers/events.router.js
```diff
@@ -1,51 +1,29 @@
...
-      console.log(err);
-      res.sendStatus(500).json({
-        message: `/GET Internal server error: ${err}`,
-      });
-    });
-});
+// Update Event by id with PATCH.
+router.patch('/:EventId', EventController.update);
+
+// TODO: Implement Get Event members by GET
```
alex-anakin: As I mentioned above we won't implement it. Please remove.

backend/routers/projects.router.js
```diff
@@ -1,81 +1,38 @@
 const express = require("express");
 const router = express.Router();
 
-const Project = require("../models/project.model");
+const { ProjectController } = require('../controllers');
+const { verifyUser } = require('../middleware');
 
-// GET /api/projects/
```
alex-anakin, line 6: Please add one top comment for all routers where you deleted comments like this `// GET /api/projects/`
For example for this router, it can be `// The root is /api/projects`.
Otherwise, it will be a little bit confusing for all newcomers or inexperienced developers.

backend/routers/projects.router.js
```diff
@@ -1,81 +1,38 @@
...
-    })
-    .catch((err) => {
-      console.log(err);
-      res.sendStatus(500).json({
-        message: `/GET Internal server error: ${err}`,
-      });
-    });
-  // };
-});
+// Create new Project with POST.
```
alex-anakin: Do we really need comments for all endpoints? As for me, they are self-explanatory.

backend/routers/projects.router.js
```diff
@@ -1,81 +1,38 @@
...
+router.get('/:ProjectId/members', [verifyUser.isAdminByEmail], ProjectController.member_list);
+
+// TODO: Implement Add a Project members with POST
+router.post('/:ProjectId/members', [verifyUser.isAdminByEmail], ProjectController.add_member);
+
+// TODO: Implement Get a Project member with GET
+router.get('/:ProjectId/members/:MemberId', [verifyUser.isAdminByEmail], ProjectController.member_by_id);
+
+// TODO: Implement Remove a Project member with POST
+router.delete('/:ProjectId/members/:MemberId', [verifyUser.isAdminByEmail], ProjectController.remove_member);
```
alex-anakin: Please remove all TODOs as we discussed changes to data models that will result in API updating.

backend/validators/auth.api.validator.js
```diff
@@ -0,0 +1,42 @@
+const { body, validationResult } = require('express-validator');
```
alex-anakin: What about file name `user.validator.js` instead of `auth.api.validator.js`? Looks like it validates not only auth but creating user calls too.

backend/validators/index.js
```diff
@@ -0,0 +1,5 @@
+const authAPIValidator = require('./auth.api.validator');
+
+module.exports = {
+  authAPIValidator,
```
alex-anakin: There are a lot of discussions on this topic https://stackoverflow.com/questions/15526107/acronyms-in-camelcase
but I personally prefer `authApiValidator` instead of `authAPIValidator`. For my eyes, it's just easier to read.

alex-anakin (changes requested):
Hi Nick, please take a look at my comments. Please use the button **Resolve conversation** where you agree with my comment and fixed the code. Otherwise, let's discuss the remaining comments at a meeting.

backend/controllers/event.controller.js
nickbeaird: 👍

backend/controllers/email.controller.js
nickbeaird: 👍

backend/routers/auth.router.js
nickbeaird: 👍

backend/routers/projects.router.js
nickbeaird: I don't think so, but I thought that I would follow with the convention that we had. I will remove them.

backend/controllers/email.controller.test.js
nickbeaird: If you can hold off on this one, then that would be stellar. I have fought with getting this one test to work for long enough.

alex-anakin: No problem, let's create a separate task for it.

alex-anakin (approved):
Looks good! Approved.

Issue #154, opened 2020-05-11
Add Slack "Greet Bot" to direct new-joiners to create a profile


sarL3y:
### Overview
Upon joining the org Slack channels, a new user should be greeted with a bot that directs them to vrms.io to create a new profile

alex-anakin:
Take a look here https://github.com/100Automations/futureautomations/issues/8

ye-susan:
_Copied from 100Automations issue 8:_ 

Goal: Replace greet.bot with a new slack bot that can pull from csv file (Google docs) and markdown files: e.g., When to send (days they have been a member), order, active(true/false), message file name
e.g.

```
    1, 0, true, welcome.md
    7, 0, true, joinedlastweek.md
    30, 0, true, active30days.md
    15, 0, false, notactive15days.md
```

when script finishes running it should report to whichever handles are in a csv file
e.g. @ bonnie, @ leader2, @ leader3, etc. as well as adding a successful or unsuccessful message to a log file in the repo.

Automation triggers (What starts it? What's it responding to?)
 -  Time-based (Specify frequency (e.g. 1x/week)
 -  Event-based (e.g. someone just created a new GitHub on a repo)
 -  The script will run once a day. Messages will be sent based on how long someone is a member and each will have a different date trigger as specified by the csv file.
 - Script will run automatically using Github Actions

**Clarifying Questions**
- What information specifically should the leaders be seeing in the log, or just a success/fail message? Will slackbot send them a message or a file?
- How does Github Actions fit into the slack-google docs-node flow? (Additional research that I need to do here)

### Action Items
- [ ] set up Slackbot to send messages to users when they join/after 7, 15, 30 days
- [ ] Determine how much manual/custom input is required for the Google doc (see how much info we can extract from Slack's Web APIs that are not paid)
- [ ] set up Google docs API - be able to read and write into it using Node
- [ ] automate data population from slackbot into the Google doc (doable?) 

Resources
https://medium.com/better-programming/make-a-team-roster-with-node-react-and-google-sheets-305b9f9ebbb1
https://api.slack.com/reference/surfaces/formatting
https://api.slack.com/start/building/bolt-js
https://api.slack.com/methods

ye-susan:
Progress: Set up Bolt and Glitch (local development not set up yet), and also reading through docs to get more familiar with the slack APIs
Blocks: reading through docs, many many docs
ETA: Hoping I can at least get bot to read user's joined date and send message accordingly by next week
[...]

PR #382, opened 2020-11-02
Slackbot to message new users

ye-susan:
Related: https://github.com/hackforla/VRMS/issues/154
I'll have to take some time off VRMS to focus on work, so I wanted to submit my PR with what I currently have.

This PR contains commits for the Slackbot to message user accounts less than 30 days old.
Currently functionality: 
* query the db for user accounts that are <= 30 days old
* message user from markdown via Slackbot direct message

Improvements Needed:
* check if user has already attended meetings, so they won't receive the messages
* this slackbot is currently not linked to cron to be run daily
* need a way to determine that we've already messaged user so they won't receive message again


alex-anakin:
@ye-susan is it possible to test it locally by a reviewer? If yes, please add steps on how to perform manual testing.

ye-susan:
Hi @alex-anakin, 
For testing locally, the reviewer would need: 
* slack tokens for their test slack workspace
* an email to test on (email should be included user table in test db and createdDate should be less than 30 days old), the email would also need to be the same email used for the slack workspace 
(since the slack API searches for the email to find the corresponding Slack ID for the slackbot to direct message)

backend/routers/slack.router.js
```diff
@@ -111,4 +114,63 @@ async function findProject(req, res) {
...
+async function findAndDmChat() {
+  try {
+    const potentialSlackUsers = await getUsers();
+    const actualSlackUsers = [];
+
+    //find users who are in our slack
+    for (let user of potentialSlackUsers) {
+      console.log("Email:", user.email)
+      try {
+        let result = await app.client.users.lookupByEmail({
```
nickbeaird, line 146: Although I know that I was the direct cause of this implementation, please take a gander at this Google best practices on async/await for slightly refactoring this section for the better. https://developers.google.com/web/fundamentals/primers/async-functions#example_outputting_fetches_in_order

I do know that you are busy with things, so just let me know if you are too busy for that.

backend/routers/slack.router.js
```diff
@@ -111,4 +114,63 @@ async function findProject(req, res) {
...
+        console.log("User is not on Slack");
+      }
+    }
+ 
+    const fileName = './assets/newMember.md';
+    const textData = fs.readFileSync(fileName,'utf8');
+    
+    //send Slack users msg from Slackbot
+    for (let slackUser of actualSlackUsers) {
+      const sendUserMsgObj = app.client.chat.postMessage({
```
nickbeaird, line 164: See the same as the above statement.

nickbeaird (changes requested):
I have not tested this manually myself, but I am throwing in a quick review on the async loops.

backend/routers/slack.router.js
ye-susan: Thanks for reviewing Nick! I will be a bit slammed in the coming weeks in terms of scheduling, maybe I can create an issue to improve on my PR? Either I can come back to it when I have more time -- or if this is blocking/urgent then someone else might take it up?

PR #404, opened 2020-11-10
Update equality tests and fixes errors.

### LINKED ISSUE #370: backend/models/test.js files are not using expect method correctly
### Overview
These test files need to use expect(VAL).toBe(VAL), which they are currently not doing.

### Action Items
REPLACE
    expect(savedData.name === submittedData.name);
WITH
    expect(savedData.name).toBe(submittedData.name);

@nickbeaird brought this up in a PR. I looked at all the test.js files for models, and they all were using expect incorrectly.

### LINKED ISSUE #417: Suppress temporarily the test causing GitHub actions failure
### Overview
The test https://github.com/hackforla/VRMS/blob/development/backend/config/auth.config.test.js fails on GitHub actions but passes locally. We need to suppress it or ignore programmatically, or just comment out. 

### Action Items

- [ ] Disable the test using any approach
- [ ] Add TODO to turn the test on after fixing GitHub actions.

awhig:


awhig:
I'm not sure why but the test is not picking up the newly added dotenv-expand package, which is causing the test to fail.

nickbeaird:
> I'm not sure why but the test is not picking up the newly added dotenv-expand package, which is causing the test to fail.



> I'm not sure why but the test is not picking up the newly added dotenv-expand package, which is causing the test to fail.

I am pretty sure that we have to explicitly add this to the secrets repo and the pipeline. Look at the Github Actions workflow, but we can pair on this at some point if you need help/access.

awhig:
@nickbeaird Ok. I'm not familiar with the github action stuff or the secrets repo, so whenever you have time. I took a look at, and I think it does a yarn install, which I would think would pull down the package...

backend/globalConfig.json
```diff
@@ -1 +1 @@
-{"mongoUri":"mongodb://127.0.0.1:42413/jest?","mongoDBName":"jest"}
\ No newline at end of file
+{"mongoUri":"mongodb://127.0.0.1:62109/jest?","mongoDBName":"jest"}
```
alex-anakin: add an empty line at the file end

alex-anakin (changes requested):
Looks good to me. Please resolve the comment and it will be good to go.
I believe we need to fix GitHub actions in a separate PR.

backend/app.js
```diff
@@ -12,7 +12,11 @@ const customRequestHeaderName = 'x-customrequired-header';
 const dontCheckCustomRequestHeaderApis = ["GET::/api/recurringevents"];
 
 // Import environment variables
-require("dotenv").config();
+const dotenv = require('dotenv');
+const dotenvExpand = require('dotenv-expand');
```
nickbeaird, line 16: 👍

backend/config/auth.config.test.js
```diff
@@ -1,4 +1,4 @@
 test('Environment variables are working as expected', () => {
   const backendUrl = process.env.REACT_APP_PROXY;
-  expect(backendUrl === `http://localhost:${process.env.BACKEND_PORT}`);
+  expect(backendUrl).toEqual(`http://localhost:${process.env.BACKEND_PORT}`);
```
nickbeaird: I was looking at toEqual and it is used for comparing objects, and not strings. This above is comparing two strings through string interpolation. I am pretty sure that is toBe() syntax instead. Both may work, and you can test this by trying to write a toEqual test that you know should fail and to see if it meets your expectations. Then rerun with a toBe() and compare. 


https://jestjs.io/docs/en/expect#toequalvalue
https://jestjs.io/docs/en/expect#tobevalue

backend/models/checkin.test.js
```diff
@@ -15,9 +15,9 @@ describe("Checkin Model saves the correct values", () => {
     await CheckIn.create(submittedCheckinData);
     const savedCheckinDataArray = await CheckIn.find();
     const savedCheckinData = savedCheckinDataArray[0];
-    expect(savedCheckinData.userId === submittedCheckinData.userId);
-    expect(savedCheckinData.eventId === submittedCheckinData.eventId);
-    expect(savedCheckinData.createdDate === submittedCheckinData.createdDate);
+    expect(savedCheckinData.userId).toEqual(submittedCheckinData.userId);
+    expect(savedCheckinData.eventId).toEqual(submittedCheckinData.eventId);
+    expect(savedCheckinData.createdDate.getTime()).toEqual(submittedCheckinData.createdDate);
```
nickbeaird: I think that you can either compare the values with toBe(), or look at comparing the object with toEqual().

nickbeaird (changes requested):

backend/models/checkin.test.js
awhig: In this case they are equivalent. I prefer to use toEqual as I don't really want to check object ref. equality.

backend/config/auth.config.test.js
awhig: In this case you can use either toBe or toEqual. I prefer to use toEqual as I'm not checking for object ref. equality.

awhig: @nickbeaird I changed to toBe

backend/models/checkin.test.js
awhig: @nickbeaird I changed to toBe

PR #405, opened 2020-11-10
Return updated record from DB

### LINKED ISSUE #403: Mongoose findOneAndUpdate must use new parameter to return updated record
### Overview
There are two places in the backend code that are using findOneAndUpdate to update and return a DB record. They return the record in the response. I assume this should be the updated record. 

By default the pre-updated record is returned, unless a "new" option is set to true. This is currently not being done.

### Action Items
Update DB call to pass "new" option as true.

### Resources/Instructions
See https://mongoosejs.com/docs/tutorials/findoneandupdate.html

awhig:


awhig:
Return the updated record from the DB.

backend/controllers/project.controller.js
```diff
@@ -38,7 +38,7 @@ ProjectController.project_by_id = async function (req, res) {
 ProjectController.update = async function (req, res) {
   const { ProjectId } = req.params;
   try {
-    const project = await Project.findOneAndUpdate(ProjectId, req.body);
+    const project = await Project.findOneAndUpdate(ProjectId, req.body, {new: true});
```
nickbeaird, line 41: 👍

nickbeaird (approved):

alex-anakin (approved):
Looks good, approved.

PR #406, opened 2020-11-10
Create ModificationLog model and Service

### LINKED ISSUE #365: Create ModificationLog data model
### Overview
`ModificationLog` stores a history of modification of any object type. We can't use `ObjectId` for a reference but using a String representation of `Id` is Ok here.

Create a Mongoose schema `ModificationLog` containing the following properties:

```
date: Date
authorEmail: String
objectId: String			// modified object id stored as String
objectType: String, enum ['UserProfile', 'EventTemplate', 'ProjectInfo', 'ProjectMembership']
newState: Object			// modified object new state should be saved completely as an object
```

### Action Items

- [x] Create a schema
- [x] Create the schema test
- [x] Define compound index using `objectId` and `objectType`

### Resources/Instructions
https://github.com/hackforla/VRMS/tree/development/backend/models

### LINKED ISSUE #388: Implement a service for operating ModificationLog entity
The task is blocked by:
- [ ] #365

### Overview
ModificationLog handling doesn't require implementing API now. We need to have a service for calling it from other services or controllers to save or get logs.

### Action Items

- [ ] Add `services/modificationLog.service.js` into `backend` folder. 
- [ ] Implement `saveLog(authorEmail, object)` function without return value.
- [ ] Implement `getLogs(objectId, objectType)` function that returns `ModificationLog` array.
- [ ] Cover implemented functionality with tests.

### Resources/Instructions
Useful articles about Node.js application architecture and using services:
https://www.coreycleary.me/project-structure-for-an-express-rest-api-when-there-is-no-standard-way
https://www.codementor.io/@evanbechtol/node-service-oriented-architecture-12vjt9zs9i
https://dev.to/santypk4/bulletproof-node-js-project-architecture-4epf#architecture

awhig:


backend/models/modificationLog.model.js
```diff
@@ -0,0 +1,33 @@
+const mongoose = require('mongoose');
+
+mongoose.Promise = global.Promise;
+
+const modificationLogSchema = mongoose.Schema({
+    date: { type: Date },   
+    authorEmail: { type: String },
+    objectId: { type: String }, // modified object id stored as String
```
nickbeaird, line 8: What is the purpose of this field? It seems like from the test that this is a primary key and an integer, but it looks to be a string here. Also, this seems redundant as we already have a n id provided by mongo. I may have some follow up depending on what approach you are taking.

nickbeaird (changes requested):

alex-anakin (approved):
Looks good! Approved.

backend/models/modificationLog.model.js
alex-anakin: It's an id of the saved object that was modified, not the id of the model.

alex-anakin (changes requested):
Sorry, after rerunning tests with applied changes the test https://github.com/hackforla/VRMS/blob/development/backend/routers/auth.router.test.js failed. Take a look at https://github.com/hackforla/VRMS/pull/406/checks?check_run_id=1403766863
We need to clean the test DB between test cases running or to use different emails.
Please fix.

awhig:
@alex-anakin I reran the test jobs, and got a different error than the previous run. I fixed those in the PR and it now passes.

nickbeaird (approved):

alex-anakin (approved):

PR #416, opened 2020-11-14
Cover LoginContainer, UserService, and EmailValidation functionality with tests

### LINKED ISSUE #400: Cover implemented LoginContainer functionality with tests
### Overview

Need to cover implemented LoginContainer functionality with unit tests (mock data, test logic).

### Resources/Instructions
Implement "Sign In" page #386

OlgaBilogurova:
Fixes #400 

Please, do `yarn install` for `client-mvp-04` because I installed new packages to mock data: "redux-mock-store", "jest-fetch-mock"


client-mvp-04/src/components/login/loginContainer.test.js
```diff
@@ -0,0 +1,102 @@
...
+jest.mock('../../services/user.service', () =>
+  jest.fn().mockImplementationOnce(() => {
+    return mockUserData;
+  })
+);
+
+// Mock history.push('/path')
+const mockHistoryPush = jest.fn();
+jest.mock('react-router-dom', () => ({
+  ...jest.requireActual('react-router-dom'),
```
nickbeaird: I think that a comment here would be helpful. It looks likes your intention is to mock the browser's history, but continue to use React's router navigation. A comment would be helpful, especially as you are mocking the module and doing a destructuring on the imports.

client-mvp-04/src/components/login/loginView.js
```diff
@@ -15,7 +15,7 @@ const LoginView = ({
   errorMsgFailedEmail,
 }) => {
   return (
-    <section data-testid="login" className="login-container">
+    <section data-testid="login-container" className="login-container">
```
nickbeaird, line 18: I really appreciate the use of id's for objects, but I wanted to see if it's necessary. Do you have a goal of grabbing this item for any of your tests now? If not, then maybe we can remove it. We only really need footholds for things we really care about.

client-mvp-04/src/components/login/loginView.test.js
```diff
@@ -25,10 +25,14 @@ describe('Login View', () => {
...
     );
+  });
 
-    expect(screen.getByTestId('login')).toBeInTheDocument();
+  test('Should display form with input and `Sign In` button', () => {
+    render(<LoginView />, { wrapper: BrowserRouter });
+    expect(screen.getByTestId('login-container')).toBeInTheDocument();
     expect(screen.getByTestId('login-form')).toBeInTheDocument();
     expect(screen.getByTestId('login-input')).toBeInTheDocument();
+    expect(screen.getByPlaceholderText('Enter your email')).toBeInTheDocument();
```
nickbeaird: Here is an example of places where you can use data-testid's, and get exactly what you want. We really care about the login form input tells the user to "enter your email". What happens when the site has two two places with this same input? These types of tests looking for a string can break down the line. You can use both tests in one to accomplish your goals. 

You can do something like data-testid="email-input", then your test can be something like expect(screen.getByTestId="email-input".toContain("Enter your email"). Then your tests are really specific to the testid and to the value that you want the page to contain. The input's CSS can change, it can move anywhere, but as long as it has the test anchor and text, then it's good to go.

nickbeaird (changes requested):

alex-anakin (approved):
Looks good! Approved.

client-mvp-04/src/components/login/loginContainer.test.js
OlgaBilogurova: Yes, here I mock the browser's history. While I don't have a test for this I'll remove mockHistoryPush.

client-mvp-04/src/components/login/loginView.js
OlgaBilogurova: I grab it to test if we have 'login-container' in the document:
https://github.com/OlgaBilogurova/VRMS/blob/802555ea92f8d394a072e2bdea9846560248b030/client-mvp-04/src/components/login/loginView.test.js#L32

client-mvp-04/src/components/login/loginView.test.js
OlgaBilogurova: 1. I have data-testid for this input and I check if it is on the page
https://github.com/OlgaBilogurova/VRMS/blob/802555ea92f8d394a072e2bdea9846560248b030/client-mvp-04/src/components/login/loginView.test.js#L34

2. Site could have two places (and even more) with this same input because loginView.test.js tests just code inside loginView.js.

2.  This line of code wouldn't work because placeholder it's not a value it's an attribute of the input.
screen.getByTestId="email-input".toContain("Enter your email")

nickbeaird (approved):

PR #426, opened 2020-11-17
Fix for client-mvp

OlgaBilogurova:
Fix import path for validation file, change import for user.service


alex-anakin:
The change is not related to failed back-end tests. Can be merged safely.

alex-anakin (approved):

PR #427, opened 2020-11-17
UserProfileService

### LINKED ISSUE #407: Implement a service for operating UserProfile entity
The task is blocked by:
- [x] #388
- [x] #410 

### Overview
For the application 3-layered structure implement a service to operate [UserProfile](https://github.com/hackforla/VRMS/blob/development/backend/models/userProfile.model.js) entity

### Action Items

- [x] Add `userProfile.service.js` into `backend/services` folder. 
- [x] Implement `createUser(userProfile, authorEmail)` function that returns the saved entity. Features:
* Minimum required information for creating a user profile is email. Validate this.
* Make a record in ModificationLogs collection.
- [x] Implement `updateUser(userProfile, authorEmail)` function that returns the updated entity. Features:
* Use Mongoose `findOneAndUpdate` with email as a key. 
* Changing the user email is not allowed. 
* Treat undefined values of fields as no-change
* Make a record in ModificationLogs collection.
- [x] Implement `getUser(id)` function that returns unique `userProfile` entity.
- [x] Implement `getUserByEmail(email)` function that returns unique `userProfile` entity.
- [x] Cover implemented functionality with tests.

Extra task: it would be great to wrap create/update actions into one transaction with saving logs but it's not required right now.

### Resources/Instructions
Useful articles about Node.js application architecture and using services:
https://www.coreycleary.me/project-structure-for-an-express-rest-api-when-there-is-no-standard-way
https://www.codementor.io/@evanbechtol/node-service-oriented-architecture-12vjt9zs9i
https://dev.to/santypk4/bulletproof-node-js-project-architecture-4epf#architecture

### LINKED ISSUE #430: Implement custom errors
### Overview
We need to use custom errors to conveniently differentiate the cause of possible issues or cases when something goes wrong.

### Action Items
- [x] Add `errors` directory into `backend` folder.
- [ ] Create the `ValidationError` class in `errors/validation.error.js` as explained in the article below.
- [x] Create the `DatabaseError` class in `errors/database.error.js` as explained in the article below.

### Resources
Example of implementation and further using https://rclayton.silvrback.com/custom-errors-in-node-js

awhig:


backend/services/userProfile.service.js
```diff
@@ -0,0 +1,69 @@
...
+const UserProfile = require('../models/userProfile.model');
+const modificationLogService = require('./modificationLog.service');
+
+const UserProfileService = {};
+
+UserProfileService.createUser = async function (userProfileData, authorEmail) {
+    
+    let newRecord = null;
+
+    const inputData = userProfileData;
```
alex-anakin: this variable is redundant

backend/services/userProfile.service.js
```diff
@@ -0,0 +1,69 @@
...
+    let updatedUserProfile = null;
+
+    const inputData = userProfileData;
+
+    // Don't allow updating of email address
+    delete inputData.signupEmail;
+
+    await mongoose.connection.transaction(async () => {
+    
+        updatedUserProfile = await UserProfile.findOneAndUpdate(signupEmail,
```
alex-anakin: specify the strict condition for the method and normalize the value: 
`signupEmail` -> `{signupEmail: signupEmail.toLowerCase()}`

backend/services/userProfile.service.js
```diff
@@ -0,0 +1,69 @@
...
+        await modificationLogService.saveLog(authorEmail, 
+            newRecord._id, "UserProfile", newRecord );          
+    });
+
+    return newRecord;
+    
+}
+
+
+UserProfileService.updateUser = async function (signupEmail, userProfileData, authorEmail) {
```
alex-anakin: Let's build this function in a way that does not allow to change the email by its design.
1. Use only `userProfileData, authorEmail` parameters
2. In this case, an updated email will mean that the mongoose doesn't find the user. So the email updating will be impossible.
3. We need to handle cases when an attempt to modify an object fails due to the absence of the object in the DB. I prefer to throw an error for this. We don't need to throw an error when we handle any `get` requests because we can just return null. But modifying should signalize that something goes wrong when a request tries to modify a non-existing object.

backend/services/userProfile.service.js
```diff
@@ -0,0 +1,69 @@
...
+
+    
+
+    return updatedUserProfile;
+    
+}
+
+UserProfileService.getUser = async function (id) {
+
+    const results = await UserProfile.findById(id);
```
alex-anakin: The returned value is unique so let's use `result` instead of `results`.

backend/services/userProfile.service.js
```diff
@@ -0,0 +1,69 @@
...
+
+UserProfileService.getUser = async function (id) {
+
+    const results = await UserProfile.findById(id); 
+    return results;
+}
+
+UserProfileService.getUserByEmail = async function (email) {
+
+    const results = await UserProfile.findOne({signupEmail: email.toLowerCase()});
```
alex-anakin: The returned value is unique so let's use `result` instead of `results`.

backend/services/userProfile.service.js
```diff
@@ -0,0 +1,69 @@
...
+    let newRecord = null;
+
+    const inputData = userProfileData;
+
+    if(Object.prototype.hasOwnProperty.call(inputData, "signupEmail") &&
+        inputData.signupEmail) {
+        inputData.signupEmail = userProfileData.signupEmail.toLowerCase();
+    }
+
+    await mongoose.connection.transaction(async () => {
```
alex-anakin: Please, wrap all mongoose calls in `try/catch` blocks. In `catch` block throw `DatabaseError` from #430 setting error message from mongoose error.
You can implement a solution for #430 in this PR.

alex-anakin (changes requested):

backend/services/userProfile.service.js
awhig: I did this in response to a lint warning. It did not like that I was modifying an input parameter. Solution was to copy it into  a method variable. I will remove this though.

backend/services/userProfile.service.js
awhig: I think I did what you requested. Please verifyl

backend/services/userProfile.service.js
awhig: Please look at new code.

backend/services/userProfile.service.js
awhig: I've wrapped the calls in try/catch and throw a DatabaseError in the catch. Because mongoose already throws a validation error when required signupEmail is not present, this gets wrapped in our custom DatabaseError class. Let me know if you want something else to happen.

backend/services/userProfile.service.js
awhig: New Code throws new custom DatabaseError if object was not found.

backend/services/userProfile.service.js
awhig: Unless you want to differentiate that error in the catch, and throw a custom ValidationError code....
8:01
we could do
catch(e if e instanceof ReferenceError) {
    // Handle reference errors here
}
8:01
and rethrow a custom ValidationError

backend/services/userProfile.service.js
```diff
@@ -0,0 +1,97 @@
...
+    let newRecord = null;
+    
+    if(Object.prototype.hasOwnProperty.call(userProfileData, "signupEmail") &&
+        userProfileData.signupEmail) {
+        userProfileData.signupEmail = userProfileData.signupEmail.toLowerCase();
+    }
+
+    await mongoose.connection.transaction(async () => {
+    
+        try
```
alex-anakin: please reformate the code from
```
try
{
// code to execute
}
catch(error)
{
// throw error
}
```
to 
```
try {
    // code to execute
} catch (error) {
    // throw error
}
```
Take a look at examples on [MDN try...catch](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch).
Btw, do you know we have the prettier installed in the project? Probably you just need to set it up.

backend/services/userProfile.service.js
```diff
@@ -0,0 +1,97 @@
...
+        userProfileData.signupEmail = userProfileData.signupEmail.toLowerCase();
+    }
+
+    await mongoose.connection.transaction(async () => {
+    
+        try 
+        {
+            newRecord = await UserProfile.create(userProfileData);
+
+            await modificationLogService.saveLog(authorEmail,
```
alex-anakin: please fix the function call format to 
```
await modificationLogService
  .saveLog(authorEmail, newRecord._id, "UserProfile", newRecord );
```

backend/services/userProfile.service.js
alex-anakin: We need to use a lot of custom validation in our application: #431 
So to not create a mess in controllers that should send responses depending on services' success or error it's better to catch all errors from other libraries and re-throw custom ones only.

backend/errors/domain.error.js
```diff
@@ -0,0 +1,13 @@
+class DomainError extends Error {
+    constructor(message) {
+      super(message);
+     // Ensure the name of this error is the same as the class name
+      this.name = this.constructor.name;
+     // This clips the constructor invocation from the stack trace.
+     // It's not absolutely essential, but it does make the stack trace a little nicer.
+     //  @see Node.js reference (bottom)
```
alex-anakin: Looks like you copied too much of the comment from the example. We don't have a reference to node.js docs here. Please remove this line.

backend/errors/domain.error.js
```diff
@@ -0,0 +1,13 @@
+class DomainError extends Error {
```
alex-anakin: Let's keep things simple and have only one level of error inheritance in our application for now.
Sorry, I didn't add some details about what part of the example from the article we need to use. 
Let's make this file `DatabaseError` and remove its inheritor.

backend/services/userProfile.service.js
```diff
@@ -0,0 +1,97 @@
...
+
+    if(Object.prototype.hasOwnProperty.call(userProfileData, "signupEmail") &&
+        userProfileData.signupEmail) 
+    {
+        userProfileData.signupEmail = userProfileData.signupEmail.toLowerCase();
+    }
+
+    await mongoose.connection.transaction(async () => {
+    
+        try
```
alex-anakin: please fix the same formatting issue here as in the previous function

backend/services/userProfile.service.js
```diff
@@ -0,0 +1,97 @@
...
+        {            
+            updatedUserProfile = await UserProfile.findOneAndUpdate(
+                {signupEmail: userProfileData.signupEmail}, 
+                userProfileData, {new: true, runValidators: true});
+    
+            if(updatedUserProfile != null)
```
alex-anakin: please use the following format
```
if (condition) {
  // code
} else {
  // code
}
```

backend/services/userProfile.service.js
```diff
@@ -0,0 +1,97 @@
...
+    return updatedUserProfile;
+
+
+  
+    
+}
+
+UserProfileService.getUser = async function (id) {
+
+    const result = await UserProfile.findById(id);
```
alex-anakin: please wrap it in `try...catch` too

backend/services/userProfile.service.js
```diff
@@ -0,0 +1,97 @@
...
+
+UserProfileService.getUser = async function (id) {
+
+    const result = await UserProfile.findById(id); 
+    return result;
+}
+
+UserProfileService.getUserByEmail = async function (email) {
+
+    const result = await UserProfile.findOne({signupEmail: email.toLowerCase()});
```
alex-anakin: please wrap it in `try...catch` too

backend/services/userProfile.service.test.js
```diff
@@ -0,0 +1,89 @@
...
+    const findUserResult = await userProfileService.getUser(result._id);
+    expect(findUserResult.signupEmail).toBe(result.signupEmail);
+
+    const findUserByEmailResult = await userProfileService.getUserByEmail(result.signupEmail);
+    expect(findUserByEmailResult.signupEmail).toBe(result.signupEmail);
+
+    done();
+  });
+
+  test("Try and create a user profile with no email. Should throw an exeception", async (done) => {
```
alex-anakin: Please replace `Try and create a user profile with no email` -> `Should not create a user profile without email`

backend/services/userProfile.service.test.js
```diff
@@ -0,0 +1,89 @@
...
+
+    await expect(userProfileService.createUser(newRecordNoEmailDefined))
+      .rejects.toThrow(DatabaseError);
+    await expect(userProfileService.createUser(newRecordEmailNull)).rejects.toThrow(DatabaseError);
+    await expect(userProfileService.createUser(newRecordNoEmail)).rejects.toThrow(DatabaseError);
+
+    done();
+  });
+
+  test("Try and update a user profile", async (done) => {
```
alex-anakin: Please rename the test from `Try and update a user profile` to `Should update a user profile with allowed values. Should not allow update email.`

alex-anakin (changes requested):

backend/errors/domain.error.js
awhig: I removed DomainError from the chain and made this file DatabaseError. I kept the inheritance to the Base Error class though, as this contains the message field, which we need and is already implemented by Node.js.

backend/services/userProfile.service.js
awhig: I used prettier, but this line was kept as is, as it was less than 100 chars. long....I manually made it this format though, as requested.

alex-anakin (approved):
Looks good! Approved.

PR #439, opened 2020-12-08
431 implement backend validation

### LINKED ISSUE #431: Implement backend validation
The task is blocked by:
- [x] #430 

### Action Items

- [x] Add `utils/validator.js` into `backend` folder. 
- [x] Implement `isNotEmpty` static function to check a String value to be defined and not null. Also check that String doesn't consist of the whitespace characters only (space, tab, no-break space, etc.). 
- [x] Implement `isSafe` static function to check a String value doesn't contain the following characters: `{`, `$`, `:`, `}`. 
- [x] Implement `isNoLonger(str, length)` static function to check a String value length to be no longer than the second parameter.
- [x] Implement `isEmail` static function to check a String value using regexp pattern from email validation https://github.com/hackforla/VRMS/blob/development/client-mvp-04/src/utils/validation.js

Every function should throw `new ValidationError('message text depending on the function')` if the check did not pass. And return nothing if the check passed.

awhig:


backend/utils/validator.js
```diff
@@ -0,0 +1,36 @@
+const ValidationError = require('../errors/validation.error');
+
+class Validator {
+  static isNotEmpty(str) {
+    if (str === null) {
+      throw new ValidationError('string is empty');
```
alex-anakin: Let's use more descriptive messages for errors. It will help a developer to recognize the issue if he didn't expect the error.
string is empty - Null string is not allowed.

backend/utils/validator.js
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    str = str.trim();
+
+    if (str === '') {
+      throw new ValidationError('string is empty');
```
alex-anakin: string is empty - Empty string is not allowed.

backend/utils/validator.js
```diff
@@ -0,0 +1,36 @@
...
+    }
+  }
+
+  static isSafe(str) {
+    if (str !== null && str.search(/[$:{}]+/) !== -1) {
+      throw new ValidationError('string contains unsafe characters');
```
alex-anakin: Please start error messages from a capital letter and insert the string value into the message to give more context of the error.

backend/utils/validator.js
```diff
@@ -0,0 +1,36 @@
...
+    }
+  }
+
+  static isNoLonger(str, length) {
+    if (str !== null && str.length > length) {
+      throw new ValidationError(`string is longer than ${length} characters`);
```
alex-anakin: Please start error messages from a capital letter and insert the string value into the message to give more context of the error.

backend/utils/validator.js
```diff
@@ -0,0 +1,36 @@
...
+    }
+  }
+
+  static isEmail(value) {
+    const pattern = /\b[a-z0-9._]+@[a-z0-9.-]+\.[a-z]{2,4}\b/i;
+    if (value.search(pattern) === -1) {
+      throw new ValidationError(`string is not valid Email address`);
```
alex-anakin: Please start error messages from a capital letter and insert the string value into the message to give more context of the error.

alex-anakin (changes requested):
In general, changes look good, let's just improve error messages.

alex-anakin (approved):
Approved

PR #515, opened 2021-03-29
487 update event template data model

### LINKED ISSUE #487: Update EventTemplate data model due to changed requerements
### Overview
Update [EventTemplate ](https://github.com/hackforla/VRMS/blob/development/backend/models/eventTemplate.model.js) model due to the [doc](https://docs.google.com/document/d/16VvC0fbaXsvEQkQaVSj8dG3wFEHgErbjJ6H3akSUoVM/edit#heading=h.aui7osumr40e).

### Action items

- [ ] Update the model
- [ ] Fix tests

akibrhast:
Closes #487 

This PR updates the event template model and the event template model test as requested in #487 following the specifications in this [doc](https://docs.google.com/document/d/16VvC0fbaXsvEQkQaVSj8dG3wFEHgErbjJ6H3akSUoVM/edit#heading=h.aui7osumr40e)

### DIFF
--- a/backend/models/eventTemplate.model.js
+++ b/backend/models/eventTemplate.model.js
@@ -5,8 +5,8 @@ mongoose.Promise = global.Promise;
 
 const eventTemplateSchema = mongoose.Schema({
   name: { type: String },
-  belongsToProjectID: { type: mongoose.Schema.Types.ObjectId }, // Active project an event is a child of, if any (structured)
-  eventManagerID: { type: mongoose.Schema.Types.ObjectId}, // ID of user that is responsible for managing event
+  belongsToProjectID: { type: String }, // Active project an event is a child of, if any (structured)
+  eventManagerID: { type: String }, // ID of user that is responsible for managing event
   locationZone: {
     type: String,
     validate: {
--- a/backend/models/eventTemplate.model.js
+++ b/backend/models/eventTemplate.model.js
@@ -6,8 +6,8 @@ mongoose.Promise = global.Promise;
 
 const eventTemplateSchema = mongoose.Schema({
     name: { type: String },
-    belongsToProjectID:{type:String}, // Ative project an event is a child of, if any (structured)
-    eventManagerID:{type:String},     // ID of user that is responsible for managing event
+    belongsToProjectID:{type:Number}, // Ative project an event is a child of, if any (structured)
+    eventManagerID:{type:Number},     // ID of user that is responsible for managing event
     locationZone: { type: mongoose.Schema.Types.ObjectId,  ref: 'TimeZone'  },                                // HfLA locations project participates at (predefined in dictionary)
     locationName:{ type: String },    // Name of event location
     description: { type: String },
@@ -37,6 +37,16 @@ eventTemplateSchema.methods.serialize = function() {
     return {
         id: this._id,
         name: this.name,
+
+        belongsToProjectID: this.belongsToProjectID,
+        eventManagerID: this.eventManagerID,
+        locationZone:this.locationZone,
+        locationName: this.locationName,
+        timeZone: this.timeZone,
+        templateID: this.templateID,
+        recurInterval: this.recurInterval,
+        monthWeek: this.monthWeek,
+
         description: this.description,
         type: this.type,
         createdDate: this.createdDate,
--- a/backend/models/eventTemplate.model.js
+++ b/backend/models/eventTemplate.model.js
@@ -38,7 +38,6 @@ const eventTemplateSchema = mongoose.Schema({
   weekDay: { type: Number, enum: [0, 1, 2, 3, 4, 5, 6] }, //day of the week for the meetings -? sunday, monday, tuesday, wednesday,thrursday, friday, saturday
   monthWeek: { type: Number, enum: [1, 2, 3, 4, 5] }, // 1st week, 2nd week, 3rd week, 4th week, 5th week - a month can have either 4 weeks or 5 weeks
   isActive: { type: Boolean },
-  project: { type: mongoose.Schema.Types.ObjectId, ref: 'Project' },
   isOnline: { type: Boolean },
   location: {
     addressLine1: { type: String },
--- a/backend/models/eventTemplate.model.js
+++ b/backend/models/eventTemplate.model.js
@@ -5,8 +5,8 @@ mongoose.Promise = global.Promise;
 
 const eventTemplateSchema = mongoose.Schema({
   name: { type: String },
-  belongsToProjectID: { type: mongoose.ObjectId }, // Active project an event is a child of, if any (structured)
-  eventManagerID: { type: Number }, // ID of user that is responsible for managing event
+  belongsToProjectID: { type: mongoose.Schema.Types.ObjectId }, // Active project an event is a child of, if any (structured)
+  eventManagerID: { type: mongoose.Schema.Types.ObjectId}, // ID of user that is responsible for managing event
   locationZone: {
     type: String,
     validate: {
--- a/backend/models/eventTemplate.model.js
+++ b/backend/models/eventTemplate.model.js
@@ -1,75 +1,89 @@
+const { Location } = require('./dictionaries/location.model');
+const { TimeZone } = require('./dictionaries/timeZone.model');
 const mongoose = require('mongoose');
-const recurIntervalEnums = ['daily','weekly','bi-weekly','monthly'];  // daily, weekly, bi-weekly, monthly 
-const monthWeekEnums = [1,2,3,4,5];                                   // 1st week, 2nd week, 3rd week, 4th week, 5th week - a month can have either 4 weeks or 5 weeks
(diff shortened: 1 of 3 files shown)


akibrhast:
~~@alex-anakin Please don't merge this even if everything does look good~~

backend/models/eventTemplate.model.js
```diff
@@ -1,20 +1,27 @@
 const mongoose = require('mongoose');
-
+const recurIntervalEnums = ['daily','weekly','bi-weekly','monthly'];  // daily, weekly, bi-weekly, monthly
```
drubgrubby: I'm not sure that making variables for all of these really buys you anything over doing it inline.  It seems nice and tidy, but it's also an extra line of code that you have to search for if you are debugging, or want to make changes.  It's a style thing, and I don't really have a problem with it, but if you are going to make variables for these, then I'd recommend making it consistent and doing a variable for 'type' as well.

backend/models/eventTemplate.model.js
```diff
@@ -1,20 +1,27 @@
 const mongoose = require('mongoose');
-
+const recurIntervalEnums = ['daily','weekly','bi-weekly','monthly'];  // daily, weekly, bi-weekly, monthly 
+const monthWeekEnums = [1,2,3,4,5];                                   // 1st week, 2nd week, 3rd week, 4th week, 5th week - a month can have either 4 weeks or 5 weeks
+const weekDayEnums = [0, 1, 2, 3, 4, 5, 6];                           // sunday,monday,tuesday, wednesday,thrursday,fridaya,saturdaya
```
drubgrubby: Typos on "fridaya, saturdaya' in the comment.

backend/models/eventTemplate.model.js
```diff
@@ -1,20 +1,27 @@
...
 mongoose.Promise = global.Promise;
 
 const eventTemplateSchema = mongoose.Schema({
     name: { type: String },
+    belongsToProjectID:{type:Number}, // Ative project an event is a child of, if any (structured)
```
drubgrubby: Typo 'Ative' in comment

backend/models/eventTemplate.model.js
```diff
@@ -1,20 +1,27 @@
...
+    eventManagerID:{type:Number},     // ID of user that is responsible for managing event
+    locationZone: { type: mongoose.Schema.Types.ObjectId,  ref: 'TimeZone'  },                                // HfLA locations project participates at (predefined in dictionary)
+    locationName:{ type: String },    // Name of event location
```
drubgrubby: This is my super-nitpickyest comment of all, but the number of spaces after the ":" on your code isn't consistent -- sometimes it's a single space, sometimes it's double.  It makes my eye keep going back to it because inconsistencies make my eye think there's a problem when I'm scanning code.  Obviously, not an functional problem, but it would make me smile if they were consistent.

backend/models/eventTemplate.test.js
```diff
@@ -7,12 +7,20 @@ describe("Event Template Model saves the correct values", () => {
   test("Save a model instance and then read from the db", async (done) => {
     const submittedData = {
       name: "eventTemplateName",
+      belongsToProjectID: 1594023390039,
+      eventManagerID: 1594023390039,
+      //locationZone: ["location1","location2"],
```
drubgrubby: Is there a reason that "locationZone" is commented out, but still here?

drubgrubby (changes requested):
@akibrhast - This looks good to me.  These are mostly nit-picky style comments.  The only one that seems important (to me) is to make the variables consistent and either use them for everything, or not at all.  

Nice work!

backend/models/eventTemplate.model.js
akibrhast: ref: should possibly be LocationZone and NOT timeZone

backend/models/eventTemplate.model.js
```diff
@@ -1,58 +1,89 @@
...
-    createdDate: { type: Date, default: Date.now },
-    weekDay: {type: Number, enum : [0, 1, 2, 3, 4, 5, 6]},
-    startTime: { type: Date },                          
-    endTime: { type: Date },                            
-    isActive: { type: Boolean},
-    project: {                                          
-        type: mongoose.Schema.Types.ObjectId,
-        ref: 'Project'
+  name: { type: String },
+  belongsToProjectID: { type: Number }, // Active project an event is a child of, if any (structured)
```
drubgrubby: I believe IDs are alphanumeric ( ex: 5eb9e2de3409ffef8194ffde) so I think this has to be type "string". ( That seems to be what other models use).

drubgrubby (dismissed):
@akibrhast - You might want to confirm my comment about the ID type, but I'm pretty sure that's correct.  Otherwise looks good.  Thanks for making the quotation marks consistent.

backend/models/eventTemplate.model.js
```diff
@@ -1,58 +1,89 @@
...
+      message: '{VALUE} is not a valid TimeZone',
     },
-    videoConferenceLink: { type: String }              
+  }, // Event’s time zone
+  templateID: { type: String }, // Event’s join ID connecting it to recurring instances for future editing
+  recurInterval: { type: String, enum: ['daily', 'weekly', 'bi-weekly', 'monthly'] }, // daily, weekly, bi-weekly, monthly
+  weekDay: { type: Number, enum: [0, 1, 2, 3, 4, 5, 6] }, //day of the week for the meetings -? sunday, monday, tuesday, wednesday,thrursday, friday, saturday
+  monthWeek: { type: Number, enum: [1, 2, 3, 4, 5] }, // 1st week, 2nd week, 3rd week, 4th week, 5th week - a month can have either 4 weeks or 5 weeks
+  isActive: { type: Boolean },
+  project: { type: mongoose.Schema.Types.ObjectId, ref: 'Project' },
```
drubgrubby: @alex-anakin -> Just curious, is "project" different from "belongsToProjectID"?

backend/models/eventTemplate.model.js
```diff
@@ -1,58 +1,89 @@
...
+    validate: {
+      validator: async function (v) {
+        let timeZone = await TimeZone.find({});
+        return timeZone[0].timeZones.indexOf(v) >= 0;
+      },
+      message: '{VALUE} is not a valid TimeZone',
     },
-    videoConferenceLink: { type: String }              
+  }, // Event’s time zone
+  templateID: { type: String }, // Event’s join ID connecting it to recurring instances for future editing
```
drubgrubby, line 36: Should this also be type "mongoose.Schema.Types.ObjectId"?

drubgrubby (commented):
Both of the comments here have to do with the type for the various ID fields. Probably wait for Alex to weigh in so we're sure that we're doing this in the right direction.

akibrhast:
Hey @alex-anakin 
I did not really understand your explanation on the difference between the type string vs type objectID
and when we should use one vs the another. Could you please clarify on the reason behind why 
- `templateID`
- `eventManagerID`
- `belongsToProjectID`

should be of `type:string` and NOT `type: mongoose.Schema.Types.ObjectId`

A string seems to lax a check on something that is a type of ID and follows certain constraints. Plus would it not be preferable to use built in validators for a mongooseID and not have to write our own validators?

### Possible Resources
- https://stackoverflow.com/questions/13850819/can-i-determine-if-a-string-is-a-mongodb-objectid/29231016#29231016
- https://mongoosejs.com/docs/schematypes.html#objectids
- https://docs.mongodb.com/manual/reference/method/ObjectId/
- https://github.com/Automattic/mongoose/issues/1959

peteplass:
Talked about in meeting, Alex to review

akibrhast:
Update from @alex-anakin via slack

> in two words, if you have ObjectIds you are always able to fetch additional data. That is the "freedom" for me. Type checking is not important for ids because even if the type is correct but the value is not you won't fetch data

` if you have ObjectIds you are always able to fetch additional data.` ... So why is this bad?

> Somebody can add a "populate" method for fetching data when it's not needed. API will return extra data every time. In some cases, there will be performance issues. On other side, frontend data handling can be more complex when API response contains redundant data. - @alex-anakin 

As such based on that answer I am changing the id instances to Strings

akibrhast:
## Progress
None since my last comment

## Blockers
busy with hfla/website

## Availability
pretty free this week

## ETA
by friday

backend/models/eventTemplate.model.js
akibrhast: been changed to string

akibrhast:
@alex-anakin this is ready for review again

backend/models/eventTemplate.model.js
alex-anakin: We don't need a reference to the project here. 
But field `belongsToProjectID` emulates a foreign key.
@akibrhast please remove the `project` field from the model.

alex-anakin (changes requested):
Please remove the `project` field from the model (see my comment to the code).

alex-anakin (approved):

PR #521, opened 2021-04-08
Added 2021

### LINKED ISSUE #520: Adding the year 2021 for new user
### Overview
We need to add the year 2021 for new users in the check in page as right now it only shows 2020.

### Action Items
- [ ] Edit calendar on check in page to include 2021

### Resources/Instructions
Check In Page
image

akibrhast:
Added 2021
closes #520

### DIFF
--- a/client/src/pages/CheckInForm.js
+++ b/client/src/pages/CheckInForm.js
@@ -51,6 +51,7 @@ const CheckInForm = props => {
     "DEC"
   ];
   const years = [
+    "2021",
     "2020",
     "2019",
     "2018",


drubgrubby:
This looks good and appears to fix the issue.  Looks good to merge.

switchzer0:
What logic options do we have to make the year choices "<= today's date[year]?" Would be nice not to hit the same snag in Jan 2022

akibrhast:
> What logic options do we have to make the year choices "<= today's date[year]?" Would be nice not to hit the same snag in Jan 2022

 IMHO, given that this change was applied to the older client. It does not need a logical option to make the year choices. Especially since we are planning to move to the newer client at some point this year.

@switchzer0

drubgrubby:
I agree with @akibrhast - and...

1) We should be sure to check that -- if this feature exists in future versions --  it automatically keeps dates current. 
2) It wouldn't be difficult to write a snippet that creates an array of this year and some number of years in the past.  But, as Akib suggested, rather then fix it, lets use it as some fire to put version .3 in the rear-view mirror.

drubgrubby (approved):

PR #536, opened 2021-04-28
Fix scrolling issue in the DEV UI KIT web page

### LINKED ISSUE #489: Fix Scrolling In The DEV UI KIT Web Page
### Overview
Currently, scrolling doesn't work properly if the screen height is small. It doesn't look like having any problem in the beginning. But if we scroll down up to the end it doesn't show all the content (depends on screen height how much content won't show).

image

### Steps to reproduce the issue
- run `npm run mvp`
- go to this url: http://localhost:3001/dev-ui-kit
- change the screen height < 825px

piall:
Fixes #489 

🌟️ **Reason**
Our `app-main` class height is changing according to the different screens but `kit-container` class height is staying **628px** across all the screens. Scrolling is happening always for **628px** height, but `app-main` height is becoming less than **628px** for some screens and the scroll bar is going down more than **app-main** height so, the scroll bar is getting disappear and UI stuck at some point.

So, adjusted **kit-container** height for different screens like **app-main** class.
|Current|This PR|
|---------|------|
| before| after|

### DIFF
--- a/client-mvp-04/src/utils/uiKit/uiKit.scss
+++ b/client-mvp-04/src/utils/uiKit/uiKit.scss
@@ -35,3 +35,21 @@
     position: static;
   }
 }
+
+@media (max-height: 900px) {
+  .kit-container {
+    height: calc(95vh - 160px);
+  }
+}
+
+@media (max-height: 896px) and (max-width: 420px) {
+  .kit-container {
+    height: calc(100vh - 160px);
+  }
+}
+
+@media (max-width: 374px) {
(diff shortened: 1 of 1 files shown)


drubgrubby:
@mdPial -> That seems to fix the problem. Nice work!  Looks ready to merge.

drubgrubby (approved):

piall:
> @mdPial -> That seems to fix the problem. Nice work! Looks ready to merge.

Thank you 😄

Issue #378, opened 2020-11-02
Fix logout button and user's name in footer


OlgaBilogurova:
### Overview

The logout button doesn't work and the user's name displayed as 'undefined'.

![Logout](https://user-images.githubusercontent.com/11681030/97906782-1afaa380-1cf9-11eb-960b-938c3769d7ce.png)

AmyMendelsohn:
## Progress
Investigating origin of issue, starting with why the user is "undefined" in "Hi, undefined: (logout)" :
1. It used to work with firebase: https://github.com/hackforla/VRMS/pull/68
2. It was broken in this PR, when firebase was removed and magic links are used instead. Specifically, the user is set to be `true` instead of the actual user with email information in hooks/useProvideAuth.js:  https://github.com/hackforla/VRMS/pull/290
## Blockers
none
## Availability
fairly flexible
## ETA
still in investigating stage

AmyMendelsohn:
### Progress
- fixed email being undefined, so now it says "Hi email@gmail.com". Fix was to send user in verifyMe request in `user.controller.js`
- attempting to force logout by creating a new verification token when logout button clicked. Not sure if this is the best approach yet.

### Blockers
- need to learn more about how to logout and how to redirect to home page

### Availability
fairly flexible

### ETA
hopefully by next week

drubgrubby:
Amy fixed the part of this issue that had to do with the name not showing up in the footer.  It was decided at the last meeting not to bother fixing the logout functionality on version 3 as it will (hopefully) not be around long enough for that to be worth the effort.  I am closing this issue.

PR #537, opened 2021-05-01
fix (user): show user name in footer

AmyMendelsohn:
This PR fixes one of the problems in #378 

User data is passed down in the response for a request to /api/auth/me. This allows the user name to be accessed in the footer.

Please note that logout functionality is not in this PR.

image

### DIFF
--- a/backend/controllers/user.controller.js
+++ b/backend/controllers/user.controller.js
@@ -170,8 +170,9 @@ UserController.verifySignIn = async function (req, res) {
   }
 };
 
-UserController.verifyMe = function (req, res) {
-  return res.sendStatus(200);
+UserController.verifyMe = async function (req, res) {
+  const user = await User.findById(req.userId);
+  return res.status(200).send(user);
 };
 
 module.exports = UserController;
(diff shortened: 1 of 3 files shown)


drubgrubby:
@AmyMendelsohn => This is really nice work, and good sleuthing to find out where to make the change.  

One thing, and I'm on the fence about even mentioning it, but...

I believe that the original issue suggested that this was supposed to show the user's name, and this is showing the email address.  Since this whole version is about to be deprecated, it doesn't seem worth spending much time on. But, if you can fix it in...I don't know, less than five minutes, we might as well do it to spec.  

But, seriously, don't spend too much time on it. If it's easy to access the username instead of the email, go for it, otherwise, we'll merge what you did and be happy about it.

AmyMendelsohn:
@drubgrubby thanks for the catch! I can make the fix, it's simple. However, I want to check that the goal is really to have the name -- the user name was never shown in that spot, it has always been the user email.

drubgrubby:
@AmyMendelsohn -> Interesting.  I guess I'd change it to their name.  "Hi" and then an email address doesn't seem as personal, plus the original issue seems to say it should be username.  Unless you feel strongly that it should be an email address, go ahead and change it to their name.  Thanks.

AmyMendelsohn:
@drubgrubby the change has been made and PR description updated!
image

drubgrubby (approved):
Looks good!  Thanks for making that change.

PR #538, opened 2021-05-01
cleanup: fix console warnings

AmyMendelsohn:
This PR fixes many of the warnings in the console when running VRMS locally. Most of the remaining warnings after this PR relate to `useEffect`.


drubgrubby (approved):
@AmyMendelsohn -> I'm going to miss those error messages. Oh, wait.  No I'm not!

Amazing that you did this...just because. 

Also, very thoughtful about some of the things you decided to leave unchanged, like variables that were used in something commented out, in case someone was going to come back to it.   Some of us would have just taken a scythe to the whole thing just to make the errors go away.  I appreciate the attention to detail you gave to this.

PR #556, opened 2021-05-06
User management page

### LINKED ISSUE #504: Implement a page for managing users
### Overview
~1. As a project manager, I want to create a new project.~ (moving to separate issue)
~2. As a project manager, I want to change my existing project details.~ (moving to separate issue)
3. As an Administrator, I want to assign/unassign managers for projects.

### Action Items

1. Login user. If the user is an admin -> it goes to the home page. We need to add a hyperlink to the project management `projects` and user management `users` pages.
2. Login user. If the user has a non-empty array `managedProjects` then redirect him to the project management page.
3. Users page. 
* Get all users. GET /api/users and GET /api/projects
* Show 2 inputs for searching users by email and by name. No users initially on the page. We have a list of users in the client already so we need just to filter it. When the admin inputs a letter we will show a filtered list of users on the page. Columns: First Name, Last Name, email, projects.
* Clicking on any user should expand an additional section with the following elements: dropdown with all projects and the "Save changes" button. The admin can choose any project and then save changes.

### Resources/Instructions
Directory for the new page https://github.com/hackforla/VRMS/tree/development/client/src/pages
https://github.com/hackforla/VRMS/blob/59ebef0b72948530ad22e572d706a2b2df4c5568/client/src/components/Navbar.js

drubgrubby:
Fixes #504 

This page allows an admin to connect users to projects so that they will be able to make changes to those projects.

It consists of: 
- A user search
- A page that allows the admin to associate (and disassociate) projects with a selected user

Notes: 
1) Rather than make two separate inputs for email and name search, I made a single input field that displays both name and email search results in different columns.  I like it, but if it's annoying to the people who are actually going to be using this I can separate the email and name searches. 

2) There was an issue with the user.controller that needed to be fixed in order to make the necessary changes to the database. 

3) The styling is very minimal, and not according to any design, so any suggestions are welcome.


alex-anakin (changes requested):
Hi David, I started your branch locally and tested the `users` page manually. I found some bugs (take a look at the attached image):
1. The "projects" link seats on the logo. Let's move the logo to the right.
2. Emails order is different from names order.
3. One email is lost but the user has it in DB (I checked it).

manual testing results

Also, can you sort users alphabetically by name? It would be really helpful.

Another thing, lines are pretty condensed and it's a little bit difficult to find out the user email without the help of a finger. Can you display the user list as a table? (see attached image).

table

Please let me know if you have any questions.

drubgrubby:
@alex-anakin -> Thanks for taking a look at the PR.  

Just a note about the search feature, as it obviously wasn't clear.  Rather than setting up one box for a name search and one box for an email search, there is one box and when you type it does BOTH an email and name search.  So, the names on the left, and the emails on the right are not linked; they are separate searches (which is why they are not lined up).  I think that will be more clear when the results are returned alphabetically.  

I can change it so it's either a name OR email search, but I liked the efficiency of this.

drubgrubby:
Per conversation with Alex - put the data results in tabs for email and name so you can switch between them, but don't get buried in data.

drubgrubby:
@alex-anakin => I have made the changes you requested: 
1) The logo no longer overlaps the navigation links
2) I have created tabs to display search results either by email or by name
3) The results have been alphabetized.

client/src/pages/UserAdmin.js
```diff
@@ -0,0 +1,318 @@
...
+                            {<ul className={searchResultType === "email" ? "search-results hide-results" : "search-results"}>    
+                                {nameResults.map((result,index) => {
+                                return (
+                                    <li key={index}>{result}</li>
+                                )})}
+                            </ul>}
+                        </div> 
+
+                        <div>
+                            {<ul className={searchResultType === "name" ? "search-results hide-results" : "search-results"}>
```
averdin2: As we discussed, I think making the searchResultType === "name" instead of "email" on line 193 and flipping where the hide-results class is added makes more sense in terms of code readability. Also on line 202, do the same witch changing "name" to "email" and flipping where hide-results class is added.

averdin2 (commented):
@drubgrubby as we discussed, I think we can make some of the styling things, like centering the components, into a separate issue. The navbar seems to be changed on other pages, so that should probably be fixed before merging. I also made a suggestion to code readability (see below).

client/src/pages/UserAdmin.js
```diff
@@ -0,0 +1,318 @@
...
+            const res = await fetch("/api/users/", {
+                headers: {
+                    "x-customrequired-header": headerToSend
+                  }
+            });
+            const resJson = await res.json();
+            setUsers(resJson);
+
+        } catch(error) {
+            alert(`fetchUsers ${error}`);
```
alex-anakin: Don't show the error to a user. Print the error to console instead. Show the following message in the alert:
"Server doesn't respond. Please refresh the page."

client/src/pages/UserAdmin.js
```diff
@@ -0,0 +1,318 @@
...
+            const res = await fetch("/api/projects/", {
+                headers: {
+                    "x-customrequired-header": headerToSend
+                    }
+            });
+            const resJson = await res.json();
+            setProjects(resJson);
+
+        } catch(error) {
+            alert(`fetchProjects: ${error}`);
```
alex-anakin: Don't show the error to a user. Print the error to console instead. Show the following message in the alert:
"Server doesn't respond. Please refresh the page."

client/src/pages/UserAdmin.js
```diff
@@ -0,0 +1,318 @@
...
+            },
+            body: JSON.stringify({ managedProjects })
+        };
+
+        try {
+            const response = await fetch(url, requestOptions); 
+            const resJson = await response.json();
+            return resJson;
+        } catch (error) {
+            console.log(`update user error: `, error);
```
alex-anakin, line 114: Add alert with the following message: "Server doesn't respond. Please try again."

client/src/pages/UserAdmin.js
```diff
@@ -0,0 +1,318 @@
...
+                </form>}
+                <div><button className="button-back" onClick={cancelEdit}>Back to search</button></div>
+            </div>
+        </div>
+    )
+};
+
+
+// Export UserAdmin
+export default UserAdmin;
```
alex-anakin: please add an empty line at the end of the file

client/src/sass/UserAdmin.scss
```diff
@@ -0,0 +1,125 @@
...
+  color: #333;
+}
+
+.hide-results {
+  display: none;
+}
+
+.light-results {
+  color: #666
+}
```
alex-anakin: please add an empty line at the end of the file.
There are many [explanations ](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline/729795) of that but basically, it's much safer to have an empty line at the end of the file if your favorite IDE doesn't add the EOL character.

alex-anakin (changes requested):

client/src/pages/UserAdmin.js
drubgrubby: code has been flipped.

client/src/pages/UserAdmin.js
drubgrubby: fixed

client/src/pages/UserAdmin.js
drubgrubby: fixed

client/src/pages/UserAdmin.js
drubgrubby: fixed

client/src/pages/UserAdmin.js
drubgrubby: done

client/src/sass/UserAdmin.scss
drubgrubby: done

alex-anakin (approved):

PR #557, opened 2021-05-06
Create Dropdown as a new UI element

piall:
Fixes #501 

This can be implemented in many ways. I have found we can implement it without using lots of JS code or any JS plugins. If we use any predefined HTML element. Because already predefined event listeners and other things are added already. If we make it fully from scratch have to make those on our own. In this case, I have used `radio input` to get the job done by wrapping it with the `label` tag.

PR #573, opened 2021-05-20
Refactoring UI-KIT class names

piall:
Previously lots of code were written already for class names with {}. So, other contributors who contributed to `uiKit.js` also followed the same style which was not necessary.
In my last PR, Ms. Olga suggested refactoring all the `class names` by replacing {} with " ", where **{}** are not necessary.

reference( https://github.com/hackforla/VRMS/pull/557#discussion_r633975937 )

### DIFF
--- a/client-mvp-04/src/utils/uiKit/uiKit.js
+++ b/client-mvp-04/src/utils/uiKit/uiKit.js
@@ -69,30 +69,26 @@ const DevUiKit = () => {
         <p className="italic">Text content</p>
         <p className="dev-comment">Main text: 'Open Sans', 14px, italic</p>
 
-        <a key={'default-link'} href={'/'}>
+        <a key="default-link" href="/">
           Link to click here
         </a>
         <p className="dev-comment">Default Link: 14px, semi-bold</p>
 
-        <a key={'create-link'} className="redirect-link accent-link" href={'/'}>
+        <a key="create-link" className="redirect-link accent-link" href="/">
           CREATE ACCOUNT
         </a>
         <p className="dev-comment">
           Accent Link: 13px, bold, $accent-royal-blue-color
         </p>
 
-        <a
-          key={'tooltip-link'}
-          className="redirect-link tooltip-link"
-          href={'/'}
-        >
+        <a key="tooltip-link" className="redirect-link tooltip-link" href="/">
           More details here
         </a>
         <p className="dev-comment">
           Tooltip Link: 10px, bold, $link-accent-color
         </p>
 
-        <a className="join-link" href={'/'}>
+        <a className="join-link" href="/">
           Join new project <span className="join-plus">+</span>
         </a>
         <p className="dev-comment">Join Link: 24px, bold</p>
@@ -107,7 +103,7 @@ const DevUiKit = () => {
         </div>
         <p className="dev-comment">Project Link: 24px, semi-bold</p>
 
-        <ErrorMessage content={'*Please enter a valid email address'} />
+        <ErrorMessage content="*Please enter a valid email address" />
         <p className="dev-comment">
           Error Message: italic, semi-bold; components/common/errorMessage
         </p>
@@ -116,14 +112,14 @@ const DevUiKit = () => {
       {/*** BUTTONS ***/}
       <div className="button-container">
         <h1 className="kit-sec-title">*** BUTTONS ***</h1>
-        <Button content={`Sign In`} />
+        <Button content="Sign In" />
         <p className="dev-comment">Primary default button, use without class</p>
-        <Button content={`Sign In`} className="btn-accent" />
+        <Button content="Sign In" className="btn-accent" />
         <p className="dev-comment">Accent button, use with class .btn-accent</p>
-        <Button content={`Sign In`} className="btn-square" />
+        <Button content="Sign In" className="btn-square" />
         <p className="dev-comment">Square button, use with class .btn-square</p>
         <button className="btn-confirm">
-          <img src={confirmIcon} className="btn-icon" alt={'confirm'} />
+          <img src={confirmIcon} className="btn-icon" alt="confirm" />
           Confirm
         </button>
         <p className="dev-comment">
@@ -141,10 +137,10 @@ const DevUiKit = () => {
             meeting details, track your contributions, and maintain a profile of
             your skills and professional development.
             <RedirectLink
-              linkKey={'footer-page'}
-              path={'/page'}
+              linkKey="footer-page"
+              path="page"
               className="tooltip-link"
-              content={` More details here.`}
+              content=" More details here."
             />
           </div>
         </div>
@@ -155,15 +151,15 @@ const DevUiKit = () => {
       <div className="inputs-container">
         <h1 className="kit-sec-title">*** INPUTS ***</h1>
 
-        <Input placeholder={'Enter your email'} type={'email'} />
+        <Input placeholder="Enter your email" type="email" />
         <p className="dev-comment">Default Input, components/common/input</p>
 
         <div className="text-field-icon-container">
-          <img src={gitHubIcon} className="text-field-icon" alt={'gitHub'} />
+          <img src={gitHubIcon} className="text-field-icon" alt="gitHub" />
           <input
             type="text"
             name="text-icon-field"
-            placeholder={`GitHub user name (not email)`}
+            placeholder="GitHub user name (not email)"
           />
         </div>
         <p className="dev-comment">Input with icon</p>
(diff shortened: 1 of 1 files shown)


drubgrubby (approved):
@mdPial - These changes look good.  Thanks for cleaning this up. 
(I'm going to leave this here for the day in case anyone else wants to review before merging).

piall:
@drubgrubby Thank you for self reviewing

client-mvp-04/src/utils/uiKit/uiKit.js
```diff
@@ -33,117 +33,107 @@ const DevUiKit = () => {
...
         <p>Text content</p>
-        <p className={'dev-comment'}>Main text: 'Open Sans', 14px, regular</p>
+        <p className="dev-comment">Main text: 'Open Sans', 14px, regular</p>
 
-        <p className={'italic'}>Text content</p>
-        <p className={'dev-comment'}>Main text: 'Open Sans', 14px, italic</p>
+        <p className="italic">Text content</p>
+        <p className="dev-comment">Main text: 'Open Sans', 14px, italic</p>
 
         <a key={'default-link'} href={'/'}>
```
averdin2: I noticed on line 72 that there are {} around the a tag's attributes like key and href. In my experience these also do not need the {} around the strings. This also occurs in various places in the code for a tags. You may want to discuss this with a more senior member.

client-mvp-04/src/utils/uiKit/uiKit.js
```diff
@@ -33,117 +33,107 @@ const DevUiKit = () => {
...
-        <Button content={`Sign In`} className={'btn-square'} />
-        <p className={'dev-comment'}>
-          Square button, use with class .btn-square
-        </p>
-        <button className={'btn-confirm'}>
-          <img src={confirmIcon} className={'btn-icon'} alt={'confirm'} />
+        <p className="dev-comment">Primary default button, use without class</p>
+        <Button content={`Sign In`} className="btn-accent" />
+        <p className="dev-comment">Accent button, use with class .btn-accent</p>
+        <Button content={`Sign In`} className="btn-square" />
```
averdin2: Where the Button component has content={\`Sign In\`}, I don't see any reason why there needs to be {} and \`\`. I think the Sign In should be a string with "" and no need for {} or \`\`. Again, perhaps ask another more senior dev about this.

client-mvp-04/src/utils/uiKit/uiKit.js
```diff
@@ -153,120 +143,120 @@ const DevUiKit = () => {
...
+        <p className="dev-comment">Tooltip, works on hover</p>
       </div>
 
       {/*** INPUTS ***/}
-      <div className={'inputs-container'}>
-        <h1 className={'kit-sec-title'}>*** INPUTS ***</h1>
+      <div className="inputs-container">
+        <h1 className="kit-sec-title">*** INPUTS ***</h1>
 
         <Input placeholder={'Enter your email'} type={'email'} />
```
averdin2: In this case, the attributes of the Input component don't need {} in my opinion. Typically, {} is used when inputting a variable or some JS functionality. {} aren't needed to input strings.

averdin2 (commented):
@mdPial I saw some other places other than className where it seems to me that {} aren't needed. I would ask a more senior dev on the team if my thinking is right, and perhaps make changes to these places as well.

piall:
@OlgaBilogurova or @drubgrubby Would you like to review all comments of @averdin2?

drubgrubby:
@mdPial -> I will review Alex's comments later today.

client-mvp-04/src/utils/uiKit/uiKit.js
drubgrubby: Agreed.  You don't need the curly braces here.

client-mvp-04/src/utils/uiKit/uiKit.js
drubgrubby: I agree, you don't need the curly braces here either.

client-mvp-04/src/utils/uiKit/uiKit.js
drubgrubby: I agree, again.

drubgrubby:
@mdPial -> The original issue was for refactoring the class names, and you did to that correctly in your original PR.  But @averdin2  made a good catch on a couple of additional instances of unnecessary curly braces that you might as well remove while you're at it. 

I agree with Alex's comments; go ahead and make those changes.

client-mvp-04/src/utils/uiKit/uiKit.js
OlgaBilogurova: I agree.

client-mvp-04/src/utils/uiKit/uiKit.js
```diff
@@ -33,117 +33,107 @@ const DevUiKit = () => {
...
+        <a key={'create-link'} className="redirect-link accent-link" href={'/'}>
           CREATE ACCOUNT
         </a>
-        <p className={'dev-comment'}>
+        <p className="dev-comment">
           Accent Link: 13px, bold, $accent-royal-blue-color
         </p>
 
         <a
           key={'tooltip-link'}
```
OlgaBilogurova: Could be replaced with: key="tooltip-link"

client-mvp-04/src/utils/uiKit/uiKit.js
OlgaBilogurova: I agree

client-mvp-04/src/utils/uiKit/uiKit.js
```diff
@@ -33,117 +33,107 @@ const DevUiKit = () => {
...
-        </p>
-        <button className={'btn-confirm'}>
-          <img src={confirmIcon} className={'btn-icon'} alt={'confirm'} />
+        <p className="dev-comment">Primary default button, use without class</p>
+        <Button content={`Sign In`} className="btn-accent" />
+        <p className="dev-comment">Accent button, use with class .btn-accent</p>
+        <Button content={`Sign In`} className="btn-square" />
+        <p className="dev-comment">Square button, use with class .btn-square</p>
+        <button className="btn-confirm">
+          <img src={confirmIcon} className="btn-icon" alt={'confirm'} />
```
OlgaBilogurova: The same for alt here.

client-mvp-04/src/utils/uiKit/uiKit.js
```diff
@@ -33,117 +33,107 @@ const DevUiKit = () => {
...
           More details here
         </a>
-        <p className={'dev-comment'}>
+        <p className="dev-comment">
           Tooltip Link: 10px, bold, $link-accent-color
         </p>
 
-        <a className={'join-link'} href={'/'}>
-          Join new project <span className={'join-plus'}>+</span>
+        <a className="join-link" href={'/'}>
```
OlgaBilogurova: The same for href.

client-mvp-04/src/utils/uiKit/uiKit.js
OlgaBilogurova: Agree.

client-mvp-04/src/utils/uiKit/uiKit.js
```diff
@@ -153,120 +143,120 @@ const DevUiKit = () => {
...
+        <h1 className="kit-sec-title">*** INPUTS ***</h1>
 
         <Input placeholder={'Enter your email'} type={'email'} />
-        <p className={'dev-comment'}>Default Input, components/common/input</p>
+        <p className="dev-comment">Default Input, components/common/input</p>
 
-        <div className={'text-field-icon-container'}>
-          <img src={gitHubIcon} className={'text-field-icon'} alt={'gitHub'} />
+        <div className="text-field-icon-container">
+          <img src={gitHubIcon} className="text-field-icon" alt={'gitHub'} />
```
OlgaBilogurova: The same for alt.

client-mvp-04/src/utils/uiKit/uiKit.js
```diff
@@ -153,120 +143,120 @@ const DevUiKit = () => {
...
 
-        <div className={'text-field-rounded-icon-container'}>
+        <div className="text-field-rounded-icon-container">
           <input
             type="text"
             name="text-icon-field"
             placeholder={`Search Reports`}
           />
-          <img src={searchIcon} className={'text-field-icon'} alt={'search'} />
+          <img src={searchIcon} className="text-field-icon" alt={'search'} />
```
OlgaBilogurova: The same for alt.

OlgaBilogurova (changes requested):
Hi @mdPial,

In all places where we pass string literals, curly braces could be removed. We just need them when passing props or variables.

Actually, both two JSX expressions are equivalent, but as for me without {} looks cleaner:
```
<MyComponent message="hello world" />
<MyComponent message={'hello world'} />
```

Please, take a look at the doc:
https://reactjs.org/docs/jsx-in-depth.html

piall:
Thank you everyone for giving the time to review this PR :smile: . I have updated.

OlgaBilogurova (approved):

PR #587, opened 2021-06-10
Add new project component

### LINKED ISSUE #560: Allow admin to create new projects
### Overview
Create a feature so that admins can create new projects

### Action Items

- [x] Add a button to the useradmin page that will open a create new projects component
- [x] Add CreateNewProject component that will allow admin to create a new project
- [x] This will only take the name of the new project.  (Additional info will be added in project management section)
- [x] Test new project name against current names, both active and inactive.

### Resources/Instructions
vrms.io/useradmin
projects collection in the db  // This is where you will see all of the fields that need to be added on new project creation.

drubgrubby:
Fixes #560 

This adds a new component to v0.3 that allows an admin to add a new project to Hack for la. 

To access new component: 
From user management page (http://localhost:3000/useradmin) click the "Add New Project" button.

The form on this page allows you to enter a new project name.
It validates the name against existing projects
if the project doesn't exist, it adds it to the database and sets it to "Active"
If it does exist, it gives you an error

While I was there, I also added a sort to alphabetize the list of projects on the add-users-to-projects component.

### DIFF
--- a/client/src/components/Navbar.js
+++ b/client/src/components/Navbar.js
@@ -37,7 +37,7 @@ const Navbar = (props) => {
                                 <p className="home-link-text">ADD EVENT</p>
                             </Link>
                             <Link to="/useradmin">
-                                <p className="home-link-text">USERS</p>
+                                <p className="home-link-text">ADMIN</p>
                             </Link>
                             <Link to="/projects">
                                 <p className="home-link-text">PROJECTS</p>
@@ -51,7 +51,7 @@ const Navbar = (props) => {
                                 <p className="home-link-text">EVENTS</p>
                             </Link>
                             <Link to="/useradmin">
-                                <p className="home-link-text">USERS</p>
+                                <p className="home-link-text">ADMIN</p>
                             </Link>
                             <Link to="/projects">
                                 <p className="home-link-text">PROJECTS</p>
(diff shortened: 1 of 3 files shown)


alex-anakin (changes requested):
Overall, the code and feature functionality look good to me.
I have just a couple of comments.
1. We definitely need to display a message to a user when a new project was successfully added to the database.
2. It would be better to place the "Add Project" button on the "Projects" page.

drubgrubby:
@alex-anakin -> I have added a notification when a new project has been successfully added. 

The following was my thinking for adding this here rather than in the projects section: 

1) Not just anyone should be able to add a new project; this functionality is reserved for Bonnie and other admins, so I put it here with the other admin feature of associating people with projects. 

2) The idea of the projects page is that people would be able to log in to manage the projects that they have been associated with in the admin feature.  If the project doesn't exist, no one can be assigned to it, so it didn't seem to make sense to add it in the projects section.

That said, I could put an "add new project" button on the "Projects" page that only displays if you are logged in as "ADMIN".  Let me know if you think that is what we should do.

drubgrubby:
@alex-anakin 
I made the requested changes
- I created a dashboard so the admin can select "manage users" or "add project"
- I added a response message for successfully adding a project to the database
- I changed the navbar title to "admin" rather than "users".  (For the moment, the url still points to 'useradmin'. 'admin' and 'admin dashboard' are already in use, and I didn't want to just go around changing names of things)

alex-anakin (approved):

PR #602, opened 2021-06-26
Fix non unique email during new profile creation checkin

akibrhast:
- Add better error message returning from backend on non-unique email
- Added help utility post function that helps making post request by passing in endpoint and object
- Update front end code to account for the new error message type [WIP]

- Delete hack nights section from newUserForm component because we dont have hacknights anymore

- Delete alredy commented out code from returnUserForm component

- Fix fetch call for successfull checkIn for retuning user only by changing `returningUser._id` to `returningUser.user._id`
	in CheckInForm.js

- Deleted unused code from CheckInForm.js

- Removed unused props that were being passed to newUserForm and ReturnUserForm

- Consolidated repeated codes checking for url params

### DIFF
--- a/backend/app.js
+++ b/backend/app.js
@@ -119,4 +119,27 @@ app.get("*", (req, res) => {
   res.sendStatus(404);
 });
 
+/** 404 handler */
+
+app.use(function (req, res, next) {
+  const err = new Error("Not Found");
+  err.status = 404;
+
+  // pass the error to the next piece of middleware
+  return next(err);
+});
+
+/** general error handler */
+
+app.use(function (err, req, res, next) {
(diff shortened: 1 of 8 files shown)


drubgrubby:
@akibrhast -> This isn't passing the backend-integration  and  backend-unit test. I tried running the tests again but it still failed.

akibrhast:
My bad. I should have noted down this is a work in progress and not ready for review.

akibrhast:
Alllll testss passing !!!

akibrhast:
I guess I could use a review on this 🗡️ 
Here is the link to the live version of this pull request.
- http://18.117.69.247/

PR #606, opened 2021-06-29
Implement controller for Location endpoints

### LINKED ISSUE #464: Implement a controller for handling Location API endpoints
The task is blocked by:
- [x] #463 

### Overview
Implement a controller for handling `Location` API endpoints.

### Action Items

- [x] Add `location.controller.js` into `backend/controllers` folder. 
- [x] Implement POST `/locations` API endpoint that returns OK 200 in the response in case of an operation success. Call `add` function of `location.service.js`. 
- [x] Implement GET `/locations` API endpoint that returns all `Location` values in the response. Call `getAll` function of location.service.js.
- [x] For all endpoints wrap the call to `location.service.js` function into `try/catch`. In case of an error get the error message and return it with relevant HTTP code (400 for validation error, 503 for database error). To avoid code duplication implement a reusable function for handling errors.
- [x] Cover implemented functionality with tests.

### Resources
The project convention about responses' HTTP codes #357

glenflorendo:
Resolves #464 

* Implement GET endpoint that retrieves all Locations

* Implement POST endpoint that creates new Location

* Fix Location services import in index

* Define status codes for validation and database errors

* Define error handling middleware

* Write tests for locations router and controller

### DIFF
--- a/backend/app.js
+++ b/backend/app.js
@@ -62,6 +62,9 @@ const runCloseCheckinWorker = require('./workers/closeCheckins')(cron, fetch);
 const runCreateRecurringEventsWorker = require('./workers/createRecurringEvents')(cron, fetch);
 // const runSlackBot = require("./workers/slackbot")(fetch);
 
+// MIDDLEWARE
+const errorhandler = require('./middleware/errorhandler.middleware');
+
 // ROUTES
 const eventsRouter = require("./routers/events.router");
 const checkInsRouter = require('./routers/checkIns.router');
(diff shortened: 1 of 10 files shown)


jasonwong26:
hi @glenflorendo 

Nice work!  I really like that you've implemented a global error handler for us.  Especially that you modified our 404 handler to leverage it!

Quick question for you: Since you've added a global error handler, what are the pros/cons of the try catch blocks in `/location.controller.js`?  Can these be removed and have us rely on the global error handler as well?

glenflorendo:
> hi @glenflorendo
> 
> Nice work! I really like that you've implemented a global error handler for us. Especially that you modified our 404 handler to leverage it!
> 
> Quick question for you: Since you've added a global error handler, what are the pros/cons of the try catch blocks in `/location.controller.js`? Can these be removed and have us rely on the global error handler as well?

So, we still need something to "route" the error to the handler. But, instead of  `try-catch` blocks, we can chain `.then`/`.catch` blocks to Promises:

```
f(...).then(...).catch(...);
```

I'll refactor my controller to reflect this syntax.

jasonwong26 (approved):
Excellent work Glen!  This all looks really good.

FYI - I doublechecked the behavior I mentioned in my last comment, and it wasn't relevant.  They added the feature in version 5, and we're running Express 4.17.1.  Sorry for the mix up!

PR #607, opened 2021-06-29
Create local storage service

### LINKED ISSUE #592: Create LocalStorage Service
## Story
Relates to feature: 1.05 - Landing/Login Page - Text Field Memory

> As a returning user, I want VRMS to remember previous inputs in the landing page text field so I don’t have to retype my email address repeatedly.

## Instructions

- [x] Review the documentation below related to localStorage.
- [x] Create a  new service for storing data to client localStorage with the following methods:
      - `getItem(key)` - retrieves the value of the specified key
      - `setItem(key, value)` - sets the value of the specified key
      - `removeItem(key)` - deletes the specified key value pair
      - `clear()` - removes all values from localStorage
- [x] Write unit tests to show you can write to, and retrieve from localStorage.

Note that localStorage converts all values to strings.  You can either have the service return the string values, or convert them to other dataTypes (e.g. parse JSON, etc.)

## Reference
- [localStorage in JavaScript: A complete guide](https://blog.logrocket.com/localstorage-javascript-complete-guide/)

piall:
Fixes #592 

This PR creates a new service for storing data to client localStorage.

### DIFF
--- a/client-mvp-04/src/services/localStorage.service.js
+++ b/client-mvp-04/src/services/localStorage.service.js
@@ -1,0 +1,53 @@
+/**
+ * Sets value in the local storage.
+ * @param {string} key - A string specifying the name of the key you want to set the value of.
+ * @param {string} value - A string specifying the value of the key.
+ */
+export function setItem(key, value) {
+  try {
+    window.localStorage.setItem(key, value);
+  } catch (err) {
+    // eslint-disable-next-line no-console
+    console.error(err);
+  }
+}
+
+/**
+ * Gets value from the local storage
+ * @param {string} key A string specifying the name of the key you want to get the value of
+ * @return {string} A string contains the value of the key
+ */
+// eslint-disable-next-line consistent-return
+export function getItem(key) {
+  try {
+    return window.localStorage.getItem(key);
+  } catch (err) {
+    // eslint-disable-next-line no-console
+    console.error(err);
+  }
+}
+
+/**
+ * Removes an item by key from the local storage.
+ * @param {string} key - A string specifying the name of the key you want to delete.
+ */
+export function removeItem(key) {
+  try {
+    window.localStorage.removeItem(key);
+  } catch (err) {
+    // eslint-disable-next-line no-console
+    console.error(err);
+  }
+}
+
+/**
+ * Clears all items from the local storage for the domain.
+ */
+export function clear() {
+  try {
+    window.localStorage.clear();
+  } catch (err) {
+    // eslint-disable-next-line no-console
+    console.error(err);
+  }
+}
--- a/client-mvp-04/src/services/localStorage.service.js
+++ b/client-mvp-04/src/services/localStorage.service.js
@@ -1,11 +1,11 @@
 /**
  * Sets value in the local storage.
  * @param {string} key - A string specifying the name of the key you want to set the value of.
- * @param {string} value - A string specifying the value of the key.
+ * @param {object | string} value - An object or a string specifying the value of the key.
  */
 export function setItem(key, value) {
   try {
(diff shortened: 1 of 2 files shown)


akibrhast (commented):
imho, 
- it would be convenient to have the functionality in `setItem` method such that I can pass an object without needing to explicitly convert it to a string.
- it would be convenient to modify have the functionality in `getItem` method such that it returns the object from the local storage and not the string.

Is there a reason why the above mentioned functionality might not be ideal?

piall:
> Is there a reason why the above mentioned functionality might not be ideal?

If I think about the input box we will need to use only string. But if I think about every case we might need to store objects in the future and object is a better option. 

**update**: I have found a new thing if we convert into JSON if we pass an object we can return an object if we pass a string we can return a string without having different implementation :smiley_cat: :exploding_head:

piall:
Appreciate the way you told me @akibrhast . Learned a new thing :smile_cat: . I have updated.

jasonwong26:
Hi @mdPial,

Excellent work on the service!  Really nice of you to include the parameter documentation as well as the unit tests.

Minor note regarding your `removes item from the local storage` test - you never set a value, so it doesn't necessarily confirm that the `removeItem()` method behaves as expected.  (But just looking at the code it certainly looks like it will!).

Couple of stylistic questions for you and the group:

- What is our policy on error handling?  I generally recommend minimizing the number of try/catch blocks in underlying services and rely on global error handlers.  But every team and project is different.
- What is the policy on writing to the console vs using an intermediary library like [debug](https://www.npmjs.com/package/debug)?  I like using that library as it is easy to toggle console messaging on and off for Dev vs Production environments.

piall:
Thanks @jasonwong26  :cat:  . As I am a junior developer (may be junior developer++ :baby:   :stuck_out_tongue_closed_eyes: ) I maybe don't do that much best practices. I am trying to learn from senior developers' approaches. Appreciate how you have brought those questions :+1:  here and also in the meeting. I have never thought in that way. Learned some new things.

piall:
>  your removes item from the local storage test - you never set a value, so it doesn't necessarily confirm that the removeItem() method behaves as expected. (But just looking at the code it certainly looks like it will!).

@jasonwong26  I didn't set any value because on the above test I set the value of **demo** key. So, I thought I can just remove that to test. If you want I can set a separate value for `removeItem()` also. 
image

piall:
> Couple of stylistic questions for you and the group:

@jasonwong26 From the opinions at the meeting. As far as I have understood we can go for it. I have never used `debug` library and `global error handlers`. I can learn and apply it.  But I felt like for the `global error handlers` gave an opinion to create a separate `issue`.  What should I do?

jasonwong26:
Hi @mdPial

Thanks for clarifying the test behavior!  Great to see that the remove test does prove the `removeItem()` method functions as expected.  If you're willing, I would encourage you to use a separate key and value for the test.  When we write unit tests we generally want them to be atomic - meaning they are not dependent on other tests.

Alternatively, you can look into Jest features for [setup and teardown](https://jestjs.io/docs/setup-teardown).  These are great tools for simplifying and managing test behaviors.  For a simple case like this, using the `beforeEach()` method is really helpful since you can have each test start with a clean slate.

Regarding the general topics that were discussed in the meeting - I'll go ahead and create some separate tickets for these.  We'll keep them completely separate from your work on this ticket.

Sound good?

piall:
@jasonwong26 
> Sound good?

Yes. If you think that is better.

piall:
I have added suggested changes.

jasonwong26 (approved):
Nice work!  approved.

piall:
Thank you @jasonwong26

PR #610, opened 2021-06-30
add label for component, and some comment in the validation

jbubar:
Made some changes to the current skills component which is already merged. @drubgrubby thanks for suggesting these changes


drubgrubby (approved):
Looks good.

PR #614, opened 2021-07-08
v.3 - Project Management Login

### LINKED ISSUE #562: Create login functionality for project management page
### Overview
When a user logs in to vrms.io - if they have a project (or projects) associated with their username - they should be sent directly to the project management page. 

They should only have access to this page and not the other admin features. 

### Action Items

- [x] Add navigation links to project management page. 
- [x] Create the above login functionality

### Resources/Instructions

TBD

drubgrubby:
Closes #562 

When a user (not admin) logs in, this allows them to log in and sends them to the project management page.  

This also fixes an issue with the positioning of "login" on the navbar

A Question: 
When a user (not admin) logs in using this functionality, there is nothing really preventing them from viewing other admin functionality.  They are never exposed to any other links (as far as I can tell), so they'd have to know it was there to use it.  In a perfect world, we would forbid access to any pages than the ones that they are authorized to see, but in this case perhaps just not being exposed to it is enough. 

Should we implement functionality that prevents non admin users from seeing other pages?  (If yes, I will create a new issue for that).


alex-anakin (approved):
LGTM

Regarding restricting access for non-admin users. For me, opening a dashboard is not criminal at all because it's just some statistic there. But the page for managing users and adding new projects should be protected from opening by a non-admin user.

PR #618, opened 2021-07-16
Create Logout Page & Implement Workflow

### LINKED ISSUE #609: Create Logout Page & Implement Workflow
## Story
Relates to feature: 2.03 Main Dashboard - Logout

> As a general user I want to log out of the product so I can keep my account secure.

## Instructions

- [x] Edit the **Action Types** file: 
  - [x] Add a new string constant named `AUTH_LOGOUT`.
- [x] Edit the **Auth Actions** file: 
  - [x] Add a new action method without parameters (you can use `authStart()` as a template), named `authLogout()` and referencing the action type you created.
  - [x] Add the new method to the exports 
- [x] Edit the **Auth Reducer** file:
  - [x] Add a new case clause to handle the new action type.
  - [x] Add logic to leave state values other than those listed below unchanged (e.g. `...state`)
  - [x] Add logic to set the following state values:
  ```
  authToken: null,
  authOrigin: null,
  loggedIn: false,
  user: null,
  userProfile: null
  ```
- [x] Add a new page component named `logout.js`
   - [x] Import the `useEffect` method from React
   - [x] Import the `useDispatch` and `useSelector` methods from react-redux
   - [x] import `allActions` from the store/actions folder
   - [x] use the useEffect method run the following on mount  (refer to the first example component as a starting point)
         - call the `dispatch()` method to trigger the logout action created above.
   - [x] add a message to the page to the effect of 'you have been logged out.' (refer to second example component for a starting point)
   - [x] create a new related test file
   - [x] add a test to confirm that the page renders the message as expected (you can use the second example page's `Should render with text content` test as an example
   - [ ] OPTIONAL: add a test to confirm that the page triggers the correct redux action on mount (see documentation on Redux testing for example)
- [x] Add the new logout page component to the Routes Collection, with the path `/logout`
- [x] Edit the Menu Component
    - [x] altering the logout link to point to the new page path
    - [x] Add a new test to confirm that clicking the Logout link redirects the user to the new logout page (use the `Should redirect to dummy page if click on "Profile" menu item` test as a template)

### Files
- **Action Types**: `/store/actions/types.js`
- **Auth Actions**: `/store/actions/authActions.js`
- **Auth Reducer**: `/store/reducers/authReducer.js`
- **Example Component with store methods**: `/components/auth/handleAuth.js`
- **Example Component for layout**: `/components/dummy/dummy.js`
- **Routes Collection**: `/routes/index.js`
- **Menu Component**: `/components/menu/menu.js`

### Libraries
- [Using the Effect Hook](https://reactjs.org/docs/hooks-effect.html)
- [Writing Redux Tests](https://redux.js.org/usage/writing-tests)

vadzimk:
Closes #609 

Not sure how to write a test "/logout page triggers the correct redux action on mount", would appreciate if you show me.
The rest of the tasks are completed.

### DIFF
--- a/client-mvp-04/src/components/logout/logout.js
+++ b/client-mvp-04/src/components/logout/logout.js
@@ -1,0 +1,22 @@
+import React, { useEffect } from 'react';
+import './logout.scss';
+import { useDispatch, useSelector } from 'react-redux';
+import allActions from '../../store/actions';
+
+const Logout = () => {
+  const loggedIn = useSelector((state) => state.auth.loggedIn);
+  const dispatch = useDispatch();
+  useEffect(() => {
+    dispatch(allActions.authActions.authLogout());
+  });
+
+  return (
+    <section data-testid="logout-message" className="flex-container dummy">
+      {!loggedIn && (
+        <h2 className="dummy-content">You have been logged out.</h2>
+      )}
+    </section>
+  );
+};
+
+export default Logout;
--- a/client-mvp-04/src/components/logout/logout.js
+++ b/client-mvp-04/src/components/logout/logout.js
@@ -8,12 +8,12 @@ const Logout = () => {
   const dispatch = useDispatch();
   useEffect(() => {
     dispatch(allActions.authActions.authLogout());
-  });
+  }, [dispatch]);
 
   return (
-    <section data-testid="logout-message" className="flex-container dummy">
+    <section data-testid="logout-message" className="flex-container logout">
       {!loggedIn && (
-        <h2 className="dummy-content">You have been logged out.</h2>
+        <h2 className="logout-content">You have been logged out.</h2>
       )}
     </section>
   );
--- a/client-mvp-04/src/components/logout/logout.scss
+++ b/client-mvp-04/src/components/logout/logout.scss
@@ -1,17 +1,17 @@
 @import "../../sass/variables";
 
-.dummy {
+.logout {
   padding: 215px 30px 100px;
   align-items: center;
 
-  .dummy-content {
+  .logout-content {
     line-height: 30px;
     font-weight: normal;
   }
 }
 
 @media (max-height: 896px), (max-width: 450px) {
-  .dummy {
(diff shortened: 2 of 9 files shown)


jasonwong26 (changes requested):
Hi Vadzim!

Really good work on this!  This was a doozy of a first issue and you really did a great job.  Two very minor change requests for you, and I've also included an example of how to test a dispatch action.

## logout.js
please add an empty array as a second parameter to the `useEffect()` call.  This will ensure the effect only fires once when the component mounts.  I've included a few links for reference as to how (and why) this works.
  - [Reference](https://reactjs.org/docs/hooks-effect.html)
  - [Relevant SO thread](https://stackoverflow.com/questions/53464595/how-to-use-componentwillmount-in-react-hooks)

```
 useEffect(() => {
    dispatch(allActions.authActions.authLogout());
  },[]); // note the empty array as a second parameter
```

## logout.scss
Looks like you've copied over the styles from `dummy.scss`.  Please assign unique class names to the styles used in your logout component so we don't have surprise effects if someone edits the dummy styles.

## logout.test.js
Your test looks really good already!  And the dispatch test was marked as optional so treat this just as an example of how you can test whether a component triggers a particular dispatch action:

```
  test('Should issue expected dispatch', () => {
    const spy = jest.spyOn(store, 'dispatch');
    render(<Logout />, { wrapper: allProviders });

    expect(spy).toHaveBeenCalledTimes(1);
    expect(spy).toHaveBeenLastCalledWith(allActions.authActions.authLogout())
  });
```
[Jest spyon documentation](https://jestjs.io/docs/jest-object#jestspyonobject-methodname)

Happy to review with you on Thursday if you want to discuss this part of it.

vadzimk:
I've done the requested changes. Thx, it was all clear to me.

jasonwong26 (approved):
Excellent work!  

Especially appreciate that you identified and included `dispatch` as a dependency for the `useEffect()` call!

PR #623, opened 2021-07-20
Add LocalStorage Service to loginContainer.js

### LINKED ISSUE #593: Add LocalStorage Service to loginContainer.js
## Story
Relates to feature: 1.05 - Landing/Login Page - Text Field Memory

> As a returning user, I want VRMS to remember previous inputs in the landing page text field so I don’t have to retype my email address repeatedly.

## Dependencies
~#592~

## Instructions

- [x] Review the storage service created for #592
- [x] Add a reference to the storage service to the `components/login/loginContainer.js` file.
- [x] Choose a key to use for storing the prior input to the landing page (you get to pick this!)
- [x] Add an internal function that uses the new service to check for an existing value in local storage for the key you chose
- [x] Add a `useEffect` call on Component Mount to call the function and set the `userEmail` state variable to this.
- [x] Modify the return section to add the `userEmail` state variable to the parameters for the `LoginView` component
- [x] Add logic to the `handleSubmitForm()` method to update the value stored in localStorage, setting it to the last submitted value.  (alternatively, only set this when the user is successfully authenticated).

### LINKED ISSUE #594: Add LocalStorage value to loginView.js
## Story
Relates to feature: 1.05 - Landing/Login Page - Text Field Memory

> As a returning user, I want VRMS to remember previous inputs in the landing page text field so I don’t have to retype my email address repeatedly.

## Dependencies
#592
#593 

## Instructions

- [ ] Review the changes made for #593, in particular the new props value being passed to `/components/login/loginView.js`
- [ ] Add the new props to the props list for this component.
- [ ] Edit the email input field to set the value to the new props value.
- [ ] Test to confirm that the following behaviors are true:
       - field updates correctly as you type
       - email value is correct on sign in
       - email value is pre-populated on next sign-in
- [ ] Update the related test class with the new prop, ensure that the tests pass.

vadzimk:
Closes #593

### DIFF
--- a/client-mvp-04/src/components/common/input/input.js
+++ b/client-mvp-04/src/components/common/input/input.js
@@ -1,6 +1,6 @@
 import React from 'react';
 
-const Input = ({ placeholder, type, onChange, autoComplete, dataTestid }) => {
+const Input = ({ placeholder, type, onChange, autoComplete, dataTestid, value }) => {
   if (!dataTestid) dataTestid = 'default-input';
   return (
     <div className="text-field-container">
@@ -11,6 +11,7 @@ const Input = ({ placeholder, type, onChange, autoComplete, dataTestid }) => {
         placeholder={placeholder}
         onChange={onChange}
         autoComplete={autoComplete}
+        value={value}
       />
     </div>
   );
--- a/client-mvp-04/src/components/login/loginContainer.js
+++ b/client-mvp-04/src/components/login/loginContainer.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import LoginView from './loginView';
 import { useDispatch } from 'react-redux';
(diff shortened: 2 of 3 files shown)


jasonwong26 (changes requested):
Hi @vadzimk,

Overall looks good!  Really appreciate you noting that we need to also set the `isDisabled` value when we load the email from local storage.

One issue - once a value is loaded from storage, users are unable edit the value.  This is because the `useEffect()` method is run each time the email input field is changed, and overwrites their entry with the value from storage.

Luckily there is an easy fix - let's change the effect to only run onMount.  Should just be a single line code change:

```
  useEffect(()=>{

     const value = getItem('vrmsuser:email');
     if(value) {
       setUserEmail(value.toString());
       setIsDisabled(false);
     }
   },[]) // empty dependencies array ensures this only runs on mount.
```

jasonwong26 (approved):
Looks good to me!  Approved.

PR #633, opened 2021-07-28
make new user email lowercase

### LINKED ISSUE #326: User email should be normalized to lowercase in database.
### Overview
During sign up, email address is stored as-is, If user types "FOO@BAR.COM", it is stored that way. Lookup is also case-sensitive, which it should not be.

### Action Items
Email should be stored as "foo@bar.com", lowercased in backend code. During lookup, any email address should be lowercased as well, in backend code.

vadzimk:
Closes #326

### DIFF
--- a/backend/controllers/user.controller.js
+++ b/backend/controllers/user.controller.js
@@ -51,8 +51,13 @@ UserController.create = async function (req, res) {
     return res.sendStatus(403);
   }
 
+  const newUser = {
+    ...req.body,
+    email: req.body.email?.toLowerCase()
+  }
+
   try {
-    const user = await User.create(req.body);
+    const user = await User.create(newUser);
     return res.status(201).send(user);
   } catch (err) {
     return res.sendStatus(400);
--- a/backend/controllers/user.controller.js
+++ b/backend/controllers/user.controller.js
@@ -51,12 +51,12 @@ UserController.create = async function (req, res) {
     return res.sendStatus(403);
   }
 
-  const newUser = {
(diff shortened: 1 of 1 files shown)


drubgrubby:
@vadzimk -> This looks like it should work just fine.  However, when I went to test it it seems that the 'check in' button that should create the new user isn't working. It isn't working for me on the development branch either, so it's not your code.  

Is the check in button working for you? 

(I remember @jasonwong26 saying that it wasn't working for him a while ago, when it was still working for me.  So...?

vadzimk:
@drubgrubby  I used the `client` instead of `client-mvp-04` to test it and it worked. The `client-mvp-04` will show 'not implemented'.

drubgrubby:
@vadzimk -> Seems like the issue is on my end as it seems to be working for Jason and Alex at the moment.  Alex said he'd take a look at this PR while I try to figure out what the issue is on my end.

jasonwong26 (changes requested):
Hi @vadzimk 

Overall, this looks really good!  One thing though - the email field is our primary key for the user model.  I think we're going to have issues we accept null values!  Please remove the optional chaining logic from your change.

```
const newUser = {
  ...req.body,
  //email: req.body.email?.toLowerCase()
  email: req.body.email.toLowerCase()
}
```

vadzimk:
HI @jasonwong26 
Done

jasonwong26 (approved):
Looks good to me!  Approved.

Issue #23, opened 2020-01-27
Add Authentication


sarL3y:
### Overview
Add Authentication. Current model uses role-based selection. We need to decide if that's the correct method for our product and pivot if needed. Ideally, we'll use a combination of magic links for users and 2FA for admins.

### Dependency
Full implementation will require new user roles once they have been added: Implement New User Roles #30 

### Requirements
* [x] Front End Developer
  * [x] Understanding of React Hooks implementation with global state management 

### Action Items
Front End:
* [x] Utilize current React Hooks and Context to implement ~Auth0~ Firebase

sarL3y:
I pivoted from Auth0, and went with Google’s Firebase. After a good amount of research and consideration, this one provided the easiest, most secure implementation for our needs. I also went with passwordless authentication for several reasons:
1) Users of this app may only be logging in once or twice a week to begin with, and remembering a password is frustrating (as well as unsecure, if repeatedly used in their personal suite of apps). By using passwords, we would have been required to use an email service for resets anyway, so this implementation removes that step.
2) Passwordless is essentially token-based, so it gives us a foundation for easily implementing social auth providers.
3) It's simple (relatively) to secure the front end with authentication being managed in global state using React Context and a Firebase class. This also means faster loading because React won’t render components that shouldn’t be shown based on access level.
4) Time to market, we’re nearing the home stretch!

Cons: 
1) I had to build a couple more pages and routes when Auth0 would have provided pre-built components to use.
2) There is a known issue with Safari that prevents Google's auth redirects from doing their job. Clearing your cache and enabling cross site tracking in Safari settings works for now. This also might be fixed by pushing to production, since Heroku adds another redirect with their temporary/subdomain apps. Info here: https://support.securly.com/hc/en-us/articles/360000881087-How-to-resolve-the-too-many-redirects-error-on-Safari- and https://support.apple.com/en-ie/HT203370

Next steps: 
1) Verify that the user is already in the database and has the proper access level (should be done by Monday but wanted to get feedback on the auth flow first). Remember this development site is using fake data, so nothing will be compromised.
2) Move Firebase project to a dedicated VRMS account. This will allow us to use a custom-formatted message when sending the links, as well as an official @vrms.io domain.
No regular members would be allowed to use this feature until member dashboards are ready (they will be turned away at the login screen because of access level), but this provides a great way to get an organizer secured right now.

vadzimk:
@sarL3y @drubgrubby I've noticed a warning that might be related
[...]

PR #654, opened 2021-08-05
redirect form /admin to /projects if user is not admin

### LINKED ISSUE #621: Restrict access to new admin dashboard for non-admin users (0.3)
### Overview
We have implemented new login functionality to allow users who have projects assigned to them to log in.  However, if they know it's there, they can still access the admin dashboard. We would like to disallow that access. 

### Action Items

The following should happen on login. 
Case 1: Person with accesslevel 'admin' logs in.  They should initially be directed to `/admin`.  

Screen Shot 2021-08-05 at 4 22 46 PM

This person should have full access to everything. 

Case 2: Person with accesslevel 'user' who has managedprojects associated with their id should be able to log in and be directed to `/projects`

Screen Shot 2021-08-05 at 4 29 47 PM

This person should ideally only have access to the `/projects` page. But, since .3 doesn't really have very robust user level authorization, they should at a minimum not have access to `/adminuser'.  

Only someone with user level 'admin' should be able to view this page: 

Screen Shot 2021-08-05 at 4 16 57 PM

The functionality to allow 'user' with managed projects to log in already exists.  Completion of this issue should restrict access of someone with accesslevel 'user'.

vadzimk:
Closes #621

@drubgrubby Not completely sure that the access to user who is not admin has been restricted bc i could not create a new user myself. 
Please check.

There is a warning that is related to useAuth hook. I commented about it in  #23

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -62,7 +62,6 @@ const App = () => {
               {routes.map(({ path, Component }) => (
                 <Route key={path} exact path={path} component={Component} />
               ))}
-              <Redirect to="/" />
             </Switch>
           </main>
           <Footer />
--- a/client/src/components/admin/dashboard/index.js
+++ b/client/src/components/admin/dashboard/index.js
@@ -329,7 +329,7 @@ const AdminDashboard = () => {
     }
   }, []);
 
-  return auth && auth.user && auth.isAdmin? (
+  return auth && auth.user && auth.user.accessLevel==='admin'? (
     <div className="flex-container">
       <div className="dashboard admin-dashboard-wrap">
 
--- a/client/src/components/admin/dashboard/index.js
+++ b/client/src/components/admin/dashboard/index.js
@@ -329,7 +329,7 @@ const AdminDashboard = () => {
     }
   }, []);
 
-  return auth && auth.user ? (
+  return auth && auth.user && auth.isAdmin? (
     <div className="flex-container">
       <div className="dashboard admin-dashboard-wrap">
 
@@ -424,7 +424,9 @@ const AdminDashboard = () => {
       </div>
     </div>
     ) : (
-    <Redirect to="/login" />
+      auth?.user
+        ?    <Redirect to="/projects" />
+        :    <Redirect to="/login" />
   );
 };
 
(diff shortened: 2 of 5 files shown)


drubgrubby (changes requested):
@vadzimk -> My apologies.  I don't think the issue itself was written very clearly.  I have re-written the issue to, hopefully, clarify the desired behavior. 

Your changes do restrict someone with access 'user' so that they can not access the `/admin/ directory, which is good. 

The directory that I'm more concerned with is the `/useradmin` directory.  This is associated with the 'Admin' navigation link, which is what confused me and why I subsequently confused you. 

Someone logged in as 'user' can still access the `/useradmin`/ directory. They should not be able to access that directory. 

Finally, when I'm logged in as a 'user' and I try to go to `/admin` it seems to kick me all the way out to the login screen, and then send me back to `/projects`.  That works, in that access is denied, but it's kind of a funky user experience. 

Nice work, and again, sorry for the lack of clarity on my part.  If you would like additional clarification, let me know and we can go over it. 

When/where are you seeing that useauth() error?  I'm not seeing it, or I'm not looking in the right place.  Perhaps we can look at it at the meeting tonight if you're going to be there.

vadzimk:
Maybe fixing authorization in this app is a bit over my head because it does not fully redirect as you probably expect.
Would be great if someone could show me the right way of solving authorization in protected routes.
I have pushed some changes. Please check.

drubgrubby (changes requested):
@vadzimk -> This is close, but not quite there.  

After this update, it does allow the correct access if you are a 'user' or an 'admin'.  However, on initial login both groups are sent back to the login page. Then when you click the 'login' button again, you go to the proper place. 

I don't have time to dig into this today, but let me look more closely on Sunday as I'm not sure exactly how to solve this either at the moment.  

Thanks for all of your work on this.

vadzimk:
@drubgrubby That redirect to '/login' happens because the user is falsy somewhere outside the components that you wanted to change access to. 
v3 uses context api to manage user state which causes a lot of trouble including that login not working properly. 
v4 is using redux which is right. If i were to touch the auth state i would change it to redux and it'd involve a lot of code changes. So i would not do it because v3 is going to be retired.

vadzimk:
@drubgrubby @jasonwong26 
In order to close it, remember there was a bug that you would need to refresh the page after clicking the magic link for the first time. Since it was fixed, now it will not redirect to login because the user is true. We just need to resolve the conflict below: `authLevelRedirect` should be called with `auth.user?.accessLevel` argument:
 `authLevelRedirect(auth.user?.accessLevel)`
I have no idea how to resolve conflicts yet.

drubgrubby:
PR incomplete. Closing until further notice.

PR #670, opened 2021-08-13
fix github action deploy vars

darpham:
Fix secrets fetching var names


jasonwong26 (approved):
Approved.  Like a ROGUE PERSON.  totally not lawful good.

PR #671, opened 2021-08-13
Manage Project Feature: Edit project info

### LINKED ISSUE #571: Edit project fields
### Dependencies
~#563~ 

### Overview
Create the functionality to edit all of the fields associated with the selected project. (This will not include meeting times). 

### Action Items
Display all fields related to selected project in editable text boxes
Write changed fields to the db

### Resources/Instructions
db.projects
vrms.io/projects
https://github.com/hackforla/VRMS/blob/development/client/src/pages/ManageProjects.js

drubgrubby:
closes #571 

By clicking the "Projects" navigation link you can select a project, and edit all of the project information stored under that project.  (This does not include meetings and other recurring events, which will be covered in the next issue). 

This is obviously not beautiful and could use some design and styling, but I believe the functionality is there.  If the team wants to spend resources on beautifying this feature after it's complete, then I suggest we do all of the design and styling at the end rather than doing it piecemeal. 

Note: The 'goFetch' function is something I was working on that I may use in refactoring some of the code in this module.  I'm just adding it in this PR even thought it's not in use yet. 

Special shout-out to @jbubar for the awesome editableField component that we used to build this.


client/src/utils/fetchUtils.js
```diff
@@ -0,0 +1,25 @@
...
+  //const headerToSend = process.env.REACT_APP_CUSTOM_REQUEST_HEADER;
+
+    const url = endPoint;
+    const requestOptions = {
+      method: fetchMethod,
+      headers: {
+          "Content-Type": "application/json",
+          "x-customrequired-header": headerToSend
+      },
+      if () {
```
jbubar: Is this empty if() statement a typo?
Wondering how it does not seem to break stuff

client/src/sass/ManageProjects.scss
```diff
@@ -17,6 +17,43 @@
...
+.project-list-heading {
+  margin-bottom: 20px;
+  font-size: 18px;
+}
+
 .project-sub-heading {
   margin: 20px 0 20px 0;
 }
+
+.editable-field {
```
jbubar: try `div .editable-field {` as a css selector.. that way you dont have to use important

backend/controllers/project.controller.js
```diff
@@ -38,7 +38,9 @@ ProjectController.project_by_id = async function (req, res) {
 ProjectController.update = async function (req, res) {
   const { ProjectId } = req.params;
   try {
-    const project = await Project.findOneAndUpdate(ProjectId, req.body, {new: true});
+    console.log('pid: ', ProjectId);
```
jbubar: Might want to remove this console.logs real quick

jbubar (changes requested):
Just a few small things that you could address, or ignore.. I might also need some updated .env variables or something? Could not get it to run because the user was underfined.
Screen Shot 2021-08-16 at 10 43 38 PM

client/src/pages/ManageProjects.js
```diff
@@ -4,18 +4,22 @@ import '../sass/ManageProjects.scss';
...
+  */
+
   const headerToSend = process.env.REACT_APP_CUSTOM_REQUEST_HEADER;
 
   const [auth] = useAuth();
   const [projects, setProjects] = useState([]);
   const [projectToEdit, setProjectToEdit] = useState([]);
   const [recurringEvents, setRecurringEvents] = useState([]);
   const [componentToDisplay, setComponentToDisplay] = useState (''); // displayProjectInfo, editMeetingTime or editProjectInfor 
-  const user = auth?.user;
```
jbubar: Could this be a mistake?

drubgrubby: Yes, that could be a mistake.  That happened when trying to resolve a conflict.  Clearly conflict resolution isn't my strength.  I've replaced that and it resolved your very first comment.

backend/controllers/project.controller.js
drubgrubby: Done.

client/src/sass/ManageProjects.scss
drubgrubby: done

client/src/utils/fetchUtils.js
drubgrubby: Very good catch.  In fact it should break stuff, and it would if I was actually using this for anything.  It's something I made for later use, and later hasn't come yet.  I have deleted this file from the PR so we don't have to entertain the existential question of the empty if statement.

drubgrubby:
@jbubar -> I believe I have addressed your concerns.  If you could take another look it would be much appreciated.

jbubar (approved):
David! Great Work!
It has been really fun to see this pr get to this point. Major Kudos!
Everything works great on the edit-project page,
But there is one really important detail. When you go back to the manage projects page and click on the same project, it shows the old unedited information..
I sent you a video in slack.
We can do this in another pr.
But needs to be fixed soon.

jbubar:
Actually, I decided to fix it real quick....
You can see in the last commit.. bb67ae7

Want to review?

drubgrubby:
@jbubar -> Thanks so much for making the fix.  I have reviewed it and it works great!  Go team!  Please merge at your earliest convenience.

PR #770, opened 2021-10-01
fix bug to make the home page scrollable

jbubar:
Fixes a bug where the home page is not scrollable.

Screen shots do not do anything because It is about scrolling..

Try going to the [dev instance here](http://dev.vrms.io/) and shrink the height, and you can see

### DIFF
--- a/client/src/pages/Home.js
+++ b/client/src/pages/Home.js
@@ -45,7 +45,6 @@ const Home = (props) => {
     }, [event]);
 
     return (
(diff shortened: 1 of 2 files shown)


drubgrubby (approved):
@jbubar -> Looks good.  I wish I could come up with something more interesting to say that "it works".  That may be the highest praise I have to offer when it comes down to it. 

Ready to merge.

PR #782, opened 2021-11-16
Fixed spacing between clickable or unclickable

### LINKED ISSUE #775: make Nav links grayed out
### Overview
Make the nav links gray out when it is on that page, to indicate that you are on that page and it is not clickable. (clickability does not actually matter)

### Action Items (possible way of doing it)
- [ ] get url params
- [ ] used url params to toggle a class
- [ ] style the class color to be gray

WalterT-MK:
closes #775 

Used conditional rendering to render nav bar clickable or not on Navbar.js
Added some css changes to Navbar.scss to fix the spacing issue.

### DIFF
--- a/client/src/components/Navbar.js
+++ b/client/src/components/Navbar.js
@@ -1,56 +1,113 @@
-import React from "react";
+import React, {useState} from "react";
 import { Link, withRouter } from "react-router-dom";
 
 import useAuth from '../hooks/useAuth';
 import {authLevelRedirect} from '../utils/authUtils'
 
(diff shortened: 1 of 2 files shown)


drubgrubby (approved):
@WalterT-MK => Really nice work!  You resolved the issue, and weren't afraid to make changes to the way it was done previously to improve the codebase.  

Ready to merge.  

My only other comment is that the title of the PR should reflect the original issue, which was about the link colors not the spacing of the links.  Of course, I was there with you when we created the PR, so that's on me, but I just want to mention it for the future.

PR #784, opened 2021-11-21
Changed checkIn form autocomplete attribute 

### LINKED ISSUE #694: Login: browser does not remember email
### Overview
As a returning user, when I login, I want the email I used previously to be remembered in the browser.

### Action Items

- [ ] Once #693 has been completed, check to see if the issue has been resolved. 
- [ ] If not, determine why the previously used email is not being remembered in the browser, or if it is remembered in some cases and not others.

chukalicious:
Closes #694 

The browser was not remembering a returning user's email address when the email input field was on focus. The auto complete attribute on the form field in the returningUser component was set to 'off' and now it's been changed to 'on. This, with the enabling of the autocomplete setting in the browser successfully shows the user's email address once the field is on focus.

### DIFF
--- a/client/src/components/presentational/returnUserForm.js
+++ b/client/src/components/presentational/returnUserForm.js
@@ -1,7 +1,6 @@
-import React from "react";
+import React from 'react';
 
 const ReturnUserForm = (props) => {
-
   return (
     <div className="check-in-container">
       <div className="check-in-headers">
@@ -10,7 +9,7 @@ const ReturnUserForm = (props) => {
       <div className="check-in-form">
         <form
           className="form-check-in"
-          autoComplete="off"
+          autoComplete="on"
           onSubmit={(e) => e.preventDefault()}
         >
           <div className="form-row">
@@ -31,9 +30,7 @@ const ReturnUserForm = (props) => {
               />
             </div>
             <p>
-              {
-                "(This allows easy use of the app. We'll never sell your data!)"
-              }
+              {"(This allows easy use of the app. We'll never sell your data!)"}
             </p>
           </div>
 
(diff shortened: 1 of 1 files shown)


drubgrubby (commented):
@chukalicious -> Good work.  I tested this in Explorer, Chrome and Opera and it seems to remember the login on all of them when I return. 

1) Sadly, but not surprisingly, it doesn't seem to work in Safari.  Can you please do a little research and see if we can make this work also in Safari, or at least understand why it doesn't seem to work in Safari?

2) This should also work on the admin login page (http://localhost:3000/login) if you are a returning user.  It seems to work for me at the moment, but if you make any additional changes for Safari, make sure that they are also working on the admin login page. 

3) When you write the description in the pull request, make the first line "closes #694", or whatever the issue number is.  If you do this, it will automatically close that issue when we merge the PR.  I added it to this PR so you can see.

chukalicious:
Hi David, thank you for reviewing! 

1.  So far: I have added name & id attributes to the input element; the label attribute is also set to email and yet the safari does not populate the info to the field. I've also found that safari only offers partial support for autocomplete. CC autocomplete info seems to work but stuff like email does not. 

- [https://cloudfour.com/thinks/autofill-what-web-devs-should-know-but-dont/](url)  

- [https://stackoverflow.com/questions/25163891/when-is-the-scan-credit-card-option-available-on-ios8-safari/36151606#36151606](url)

2. Below is a capture of the /login route remembering previously entered values. Please check on your end and let me know if this is working 😊 
image

3. Thank you for that pointer. I've never really worked with Github issues before so I appreciate the tips 😉

drubgrubby (approved):
@chukalicious -> Thanks for your good work on this.  Ready to merge, and merge I shall.

PR #787, opened 2021-11-30
fix bug such that the user can get to success page on checkin

### LINKED ISSUE #785: Redirect returning user to success screen after successful check-in
### Overview
A returning user is not being redirected to the success screen after successfully checking in
In the checkinForm.js, an API request was being made to an endpoint that does not return a response with a valid user user id

### Action Items
- [ ] The endpoint was modified to get the part of the response that returns a user id

jbubar:
fixes #785 

Had a great meeting!
Yall missed out if you were not there!

turns out we just needed to key into the returning user object..


drubgrubby (approved):
I was there!  I saw it happen!  It was amazing!  (And it works).

Nice work everyone!

PR #1057, opened 2022-02-12
Handle existing user on registration

### LINKED ISSUE #783: Missing error handling: App breaks when an already existing user tries to register again
### Overview
Missing error handling: The app breaks when an already registered user tries to register again. It seems there is no error handling in several parts of our application

### Action Items

 - [ ] understand how errors are currently being handled in the userController.js
 - [ ] confirm if the user already exists and what response is sent
 - [ ] modify the response to get more useful information on the error
 - [ ] display error in the front end

### Resources/Instructions
https://levelup.gitconnected.com/handling-errors-in-mongoose-express-for-display-in-react-d966287f573b

glenflorendo:
Fixes #783

### DIFF
--- a/backend/controllers/user.controller.js
+++ b/backend/controllers/user.controller.js
@@ -59,7 +59,13 @@ UserController.create = async function (req, res) {
   }
     const user = await User.create(newUser);
     return res.status(201).send(user);
-  } catch (err) {
+  } catch (error) {
+    if (error.name === 'MongoError' && error.code === 11000) {
+      return res.status(409).json({
+        error,
+        message: 'User already exists',
+      });
+    }
     return res.sendStatus(400);
   }
 };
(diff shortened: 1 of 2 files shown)


drubgrubby (changes requested):
@glenflorendo -> 

The app doesn't seem to break when an existing user sets up a new profile, which is good.  However, it doesn't seem to prevent me from creating duplicate users with the same email, and it never seems to display the "User already exists" error message. 

From the check-in page I clicked "Create New Profile" and then added another user using the same email address as I usually use and it created a new account without any kind of feedback.  

Screen Shot 2022-02-13 at 4 29 21 PM


Was this intended to prevent duplicate account creation, or to just keep the system from breaking when someone tries?

glenflorendo:
> 

Hi, @drubgrubby! The code change was just to handle the error when users try to register with an email address that's already in the system. From my understanding, MongoDB should not create another account with an existing email address, since the `email` field is a unique index. At least, that is the case for this specific collection: `vrms-test > users`.

drubgrubby (approved):
This is all good. We will need to make sure that the production database has unique email address.

PR #1075, opened 2022-02-22
Add 'REMOTE' to location elements if not specified.

### LINKED ISSUE #273: Attempting to generate a new events from the recurring events collection will cause an exception error if a document does not have a location property.

evanyang1:
That way, location always exists. Fixes #273

### DIFF
--- a/backend/workers/createRecurringEvents.js
+++ b/backend/workers/createRecurringEvents.js
@@ -93,9 +93,9 @@ module.exports = (cron, fetch) => {
                     } 
                     if (filteredEvent.hasOwnProperty("location")) {
                         eventToCreate.location = {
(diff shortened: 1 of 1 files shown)


drubgrubby (approved):
@evanyang1 -> Nice work.  That seems like a good solution to the problem.  Ready to merge.

PR #1079, opened 2022-03-09
alert when event is created

### LINKED ISSUE #1042: Create new event: User Feedback
### Overview
When a user creates a new recurring event they should get some feedback that their event has been successfully created.   

### Action Items
- [ ] When a user creates a new event, return them to the Project Edit page and provide feedback:  "Event Created"  

### Acceptance Criteria
When  a user creates a new event they will be redirected to the Project Edit page and get feedback so they know their event was successfully created. 

### Resources/Instructions
Create New Event
client/src/components/manageProjects/createNewEvent.js

Project Edit Page
client/src/components/manageProjects/editProject.js

evanyang1:
This resolves #1042 
Had to update react to 17 because of an error 
https://stackoverflow.com/questions/65527359/got-cant-resolve-react-jsx-runtime-error-while-use-try-to-create-the-shared-c


jbubar (commented):
In the past on this team they requested that I do not add npm modules to the project unless it is absolutely necessary. Do you think you can do this without adding boostrap. react-bootstrap and jsx-runtime?

Also, updating react does not sound like a bad thing to me, but I would defer to someone else on that.

FoxShaunR:
> In the past on this team they requested that I do not add npm modules to the project unless it is absolutely necessary. Do you think you can do this without adding boostrap. react-bootstrap and jsx-runtime?
> 
> Also, updating react does not sound like a bad thing to me, but I would defer to someone else on that.

I'm also not a big fan of adding react-bootstrap for this, especially since we're only using an Alert component. It would be better to build our own (even if it isn't as full-featured as what they offer). Updating to react 17 just for one package isn't ideal either.

Issue #1044, opened 2022-01-31
Delete Recurring Event: User Feedback


drubgrubby:
### Overview
When a user deleted a recurring event they should receive some feedback that the delete has been successful. 

### Action Items
-[ ] When the user deletes an event, provide some feedback that their event has been deleted. "Event Deleted". 

### Acceptance Criteria
When a user deletes an event, they will get feedback so they know their event has been deleted. 

### Resources/Instructions
The delete functionality is part of the update event component

Editable Meeting Component
client/src/components/manageProjects/editableMeeting.js

<details>
<summary>Delete Event</summary>
<img width="499" alt="Screen Shot 2022-01-31 at 2 07 10 PM" src="https://user-images.githubusercontent.com/68244054/151883394-0d581512-d489-4780-807f-e9862e83c5e7.png">
</details>

Issue #1043, opened 2022-01-31
Update recurring event: user feedback


drubgrubby:
### Overview
When a user updates a recurring event they should receive some feedback that the update has been successful. 

### Action Items
-[ ] When the user clicks the "update" button provide feedback that their update has been successfully competed.  Ex. "Update Successful"
-[ ] They should remain on the update screen, in case they have further updates.  There should be a link to return to the project edit page. 

### Acceptance Criteria
When the user updates an event they receive feedback that they have successfully updated an event. 

### Resources/Instructions
Project Edit Page
client/src/components/manageProjects/editProject.js

Editable Meeting Component
client/src/components/manageProjects/editableMeeting.js

PR #1081, opened 2022-03-15
user feedback when adding, editing, and deleting events

evanyang1:
solves #1042 and #1043 and #1044 (might update this more)


client/src/components/manageProjects/editableMeeting.js
```diff
@@ -15,6 +15,7 @@ const EditableMeeting = ({
   handleEventDelete,
   formErrors,
   videoConferenceLink = '',
+  setShowUpdateEventAlert,
```
FoxShaunR: Is this being used for anything?

client/src/components/manageProjects/editProject.js
```diff
@@ -155,6 +163,9 @@ const EditProject = ({
       />
       <div className="event-list">
         <h3>Recurring Events</h3>
+        {showUpdateEventAlert}
```
FoxShaunR: Looks like you're showing the update event alert twice. Here and on line 168

client/src/components/manageProjects/editProject.js
```diff
@@ -24,6 +24,12 @@ const EditProject = ({
   const [selectedEvent, setSelectedEvent] = useState();
   const [isCreateNew, setIsCreateNew] = useState();
 
+  // States for alerts
+  const [showNewEventCreatedAlert, setShowNewEventCreatedAlert] = useState(
```
FoxShaunR: What if we just have a single alert state? `const [eventAlert, setEventAlert] = useState(null);` You can then just call setEventAlert('Event Updated!') from the child components with the alert text.

FoxShaunR:
Looks pretty good. Only issue is that, if the user doesn't have the top of the event list scrolled into view, they can't see the alert. What if we position it at the bottom of the screen so that it's always visible?
```
.event-alert {
  position: fixed;
  bottom: 20px;
  right: 20px;
  background-color: white;
}
```

FoxShaunR (approved):
Looks great, thanks!

PR #1083, opened 2022-03-18
docs: add env example files

FoxShaunR:
It's standard practice to include an .env.example file when using env files. It shouldn't include sensitive information, but including things like ports and other settings for local debug is fine.  This example should be updated whenever somebody starts using a new env value.

I suspect the reasons we're getting that recurring events error is because we aren't using a consistent name for CUSTOM_REQUEST_HEADER on the backend. Including this example file should help reduce similar problems in the future


drubgrubby (approved):
Great PR.  Every code base needs a Rosetta Stone.

Issue #1080, opened 2022-03-11
Stay logged in for longer


jbubar:
### Overview
As an admin you have to log in using the magic link (email), and then 10 minutes after you have to log in again. Could we investigate why this is happening and make it so that the admin can be logged in for longer?

### Action Items
- [ ] investigate why we get signed out
- [ ] discover a solution to have the browser remember the user for longer

FoxShaunR:
Closing since we're going to move to v.4. Still nice to have for v.3 if Tim decides to reopen.

PR #1084, opened 2022-03-19
Increase token lifetime

tmshkr:
Issue: https://github.com/hackforla/VRMS/issues/1080

* Increases token lifetime to 7 days
* Adds `JWT_SECRET` environment variable (make sure to set on Heroku before merging)


backend/config/auth.config.js
```diff
@@ -1,8 +1,7 @@
 /*eslint-disable */
 module.exports = {
-  SECRET:
-    'c0d7d0716e4cecffe9dcc77ff90476d98f5aace08ea40f5516bd982b06401021191f0f24cd6759f7d8ca41b64f68d0b3ad19417453bddfd1dbe8fcb197245079',
+  SECRET: process.env.JWT_SECRET,
```
FoxShaunR, line 3: How would you feel about pulling out this secret fix into another PR and getting the token expiration fix in now?

FoxShaunR: Alright, dug into the actions some more. We do need to set this value for production before we merge, but to get your unit tests to pass (and to be in-line with a check we have for env values), do the following:

- Add  `envkey_JWT_SECRET ` to .github/workflows/all-PRs.yaml beneath line 47 with a dummy secret (we'll need to reference the github action secret once we add it)
- Add JWT_SECRET to the `require('assert-env')([...])` array in backend/src/app.js

jbubar:
@ExperimentsInHonesty We need your help to add a github secret for this new environment variable...
We want to send the secret from github to the AWS instance.

backend/config/auth.config.js
FoxShaunR: JWT_SECRET has been added to the github secrets in the dev environment.

jbubar:
@tmshkr this has been unblocked, and should be good to go if you address shaun's comments

FoxShaunR:
Closing for now. Can be reopened if Tim would like to finish this up

PR #1092, opened 2022-03-28
1045 Dynamically name recurring events

### LINKED ISSUE #1045: Dynamically name recurring events
### Overview
The recurring event "name" field is now being filled with boilerplate, not very useful, information. We want to create a dynamic naming of the event.  This will provide consistency, and reduce mistakes. 

### Action Items
- [ ] The 'create new event' component has a drop-down selector for 'Meeting Type' with the options of 'Team Meeting' and 'Onboarding'.  If the user selects "Team Meeting" they will be presented with a second drop-down with the following: 
All Team
Data Science
UI/UX Design
UI/UX Design & Research
UI/UX Research
Engineering All
Engineering Front End
Engineering Back End
PM
PM & Org
Other
- [ ] Once those two selections are made, create logic to fill the "name" and "description" fields.  The logic will be something like: 
Team Meeting -> All Team = VRMS All Team meeting 
Food Oasis -> Engineering All = Food Oasis Engineering All
- [ ] Remove the ability to edit the event name directly. The name can only be filled in the manner described above. 
- [ ] Add validation so that the new event cannot be created until both of these drop-downs have been selected. 

### Acceptance Criteria
The 'name' field in the event will be filled dynamically using dropdowns and business logic described above. 

### Resources/Instructions
Create New Event
client/src/components/manageProjects/createNewEvent.js

Note: The description field is currently being used as the name of the event for the display on the HfLA site: https://www.hackforla.org/events/.  We will currently fill the description field with the name of the meeting, but ultimately this field will be used to add further description to the meeting, e.g. Meets alternate Tuesdays

evanyang1:
maybe i need to work on the css but from the logic point of view, fixes #1045

### DIFF
--- a/client/src/components/manageProjects/createNewEvent.js
+++ b/client/src/components/manageProjects/createNewEvent.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import '../../sass/ManageProjects.scss';
 import { findNextOccuranceOfDay } from './utilities/findNextDayOccuranceOfDay';
 import { addDurationToTime } from './utilities/addDurationToTime';
@@ -14,8 +14,9 @@ const CreateNewEvent = ({
 }) => {
   // These are the initial form values
   const initialFormValues = {
-    name: `${projectName} Team Meeting`,
-    eventType: 'Team Meeting',
+    name: '',
+    eventType: '',
+    eventSubtype: '',
     description: '',
     videoConferenceLink: '',
     day: '0',
@@ -29,6 +30,11 @@ const CreateNewEvent = ({
   const handleInputChange = (event) => {
     setFormValues({ ...formValues, [event.target.name]: event.target.value });
   };
+
+  useEffect(() => {
+    let type = formValues.eventType, subtype = formValues.eventSubtype;
+    setFormValues({...formValues, name: type + " " + subtype});
+  }, [formValues.eventSubtype, formValues.eventType])
 
   const handleEventCreate = () => {
     const date = findNextOccuranceOfDay(formValues.day);
@@ -54,6 +60,7 @@ const CreateNewEvent = ({
       hacknight: 'Online',
       brigade: 'Hack for LA',
       eventType: formValues.eventType,
+      eventSubtype: formValues.eventSubtype,
       description: formValues.description,
       project: projectID,
       date: startDateTimeGMT,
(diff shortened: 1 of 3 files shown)


FoxShaunR:
Evan will make css changes in a few PR. This is ready for review.

client/src/components/manageProjects/eventForm.js
```diff
@@ -11,19 +11,25 @@ const EventForm = ({
...
         Event Name:
-        <input
-          id="name"
-          placeholder="Meeting name..."
-          name="name"
-          value={formValues.name}
-          onChange={handleInputChange}
-          maxLength={30}
-        />
+        <p>{formValues.eventType} {formValues.eventSubtype || ""}</p>
```
FoxShaunR: Use formValues.name here since it will be updated right away when eventType or eventSubtype changes. That way if we change how we're building the name in the future we don't need to update the logic here and the useEffect in createNewEvent

client/src/components/manageProjects/createNewEvent.js
```diff
@@ -103,6 +110,10 @@ const CreateNewEvent = ({
             type="button"
             className="create-form-button"
             onClick={() => {
+              if (formValues.eventType == null) {
```
FoxShaunR, line 118: Instead of displaying an alert. Take advantage of the formErrors object we're building and display an error message below the input on the eventForm.

I know we do it all over the place, but using an alert is an unpleasant user experience (can't style them, can only show one at a time, blocks other actions on the page, etc.). We should try to find better ways to show errors in our app moving forward.

FoxShaunR (changes requested):
Looking great so far. Just a couple small changes.

client/src/components/manageProjects/createNewEvent.js
```diff
@@ -116,7 +116,7 @@ const CreateNewEvent = ({
             className="create-form-button"
             onClick={() => {
               if (formValues.eventType == null) {
-                alert("Please select an event type.");
+                setFormErrors("Event type is blank. Select an event type.")
```
FoxShaunR, line 119: This should be setting errors to an object with the key being the field name and the value being the error message:

`setFormErrors({ eventType: "Event type is blank. Select an event type." })`

FoxShaunR: This is still not fixed.

jbubar:
@evanyang1 did you want to finish this or do we want to close this pr?

jbubar:
Closing this because it still needs work and is not super relevant

PR #1112, opened 2022-04-11
validate event links

### LINKED ISSUE #1106: Add validation to team meeting in formation to make sure it is a real link - Regex??
### Overview
When someone creates a new meeting, we should have some regex validation for their meeting link.

### Action Items
- [ ] find a good regex for zoom links
- [ ] implement it on the front end such that they cannot submit the post request  if it does not match the link format
- [ ] create an error message

evanyang1:
fixes #1106


jbubar (commented):
Just tested this out and it looks great! Would be happy to approve. one thing I might change is make it so thqt they need a meeting identifier after the '/' in https://us06web.zoom.us/.... you need this part at the end https://us06web.zoom.us/**j/88405283574**


Screen Recording 2022-04-21 at 6 13 41 PM

FoxShaunR:
We should have a regexp that accepts the following as valid:

www.sometext.com
https://www.sometext.com
sometext.net

and other similar variations.

FoxShaunR (approved):

PR #1114, opened 2022-04-18
Add Events Page Edit Button for Each Event

evanyang1:


jbubar (changes requested):
oooh Getting fancy here!
Screen Shot 2022-04-21 at 6 07 33 PM


This might be a bit complicated, but is it possible to redirect to the edit event modal when this button is pushed?

Also, I think we might have to rethink the styling, but that is less important rn

FoxShaunR:
Created https://github.com/hackforla/VRMS/pull/1150 to finish this change up

PR #1140, opened 2022-06-06
Fix/cannot use enter key to log in

### LINKED ISSUE #1118: Use the Enter Key to sign in
### Overview
Currently, if you click enter after typing in your email, it does not sign you in, you have to click the 'LOGIN' button

### Action Items
- [ ] figure out what happens when the button is clicked
- [ ] figure out what happens when the enter is pushed
- [ ] fix the logic for the enter
- [ ] maybe add some regex validation if there is not one

chukalicious:
Resolves #1118 

Created `useEffect` function to handle a keydown event for the 'enter' key that triggers the `handleLogin()` function created for the component, allowing the user to reach the Success page by pressing the 'enter' key, as well as the form's submit button.

### DIFF
--- a/client/src/components/auth/Auth.js
+++ b/client/src/components/auth/Auth.js
@@ -79,7 +79,6 @@ const Auth = () => {
     const inputValue = e.currentTarget.value.toString();
     validateEmail();
     if (!inputValue) {
-      console.log('inputValue inside the if statement: ', inputValue);
       setIsDisabled(true);
       showError('Please enter a valid email address');
     } else {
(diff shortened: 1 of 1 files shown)


client/src/components/auth/Auth.js
```diff
@@ -73,7 +77,9 @@ const Auth = () => {
 
   function handleInputChange(e) {
     const inputValue = e.currentTarget.value.toString();
+    validateEmail();
     if (!inputValue) {
+      console.log('inputValue inside the if statement: ', inputValue);
```
evanyang1, line 82: If I recall correctly console logs shouldn't be in the final code

evanyang1: If I recall correctly we don't want console logs in the codebase.

```suggestion
      
```

evanyang1 (approved):
LGTM! Do want someone else to check it first before merging

client/src/components/auth/Auth.js
chukalicious: Thanks Evan! I didn't see that console.log there. I took it out.

PR #1150, opened 2022-06-26
feat: events page edit button

### LINKED ISSUE #1109: Add Events Page Edit Button for Each Event
### Overview
Add Events Page Edit Button for Each Event

### Action Items
- [ ] Add an edit button to each event (may be a separate issue)
    - [x] Pass a query parameter to project page indicating project id to open
    - [x] Update project page to check for query parameter and open corresponding project

FoxShaunR:
Adds an edit button to the events page that takes the user to the corresponding project page where they can access the recurring events list to make changes. Closes: https://github.com/hackforla/VRMS/issues/1109

Also tweaked styling in a few places and added a translucent overlay to the edit and add recurring events form to prevent accidental clicks on the project form while the event form is open.

### DIFF
--- a/client/src/pages/Event.js
+++ b/client/src/pages/Event.js
@@ -1,82 +1,95 @@
 import React, { useState, useEffect } from 'react';
 import { Link } from 'react-router-dom';
 import moment from 'moment';
-import { REACT_APP_CUSTOM_REQUEST_HEADER } from "../utils/globalSettings";
+import { REACT_APP_CUSTOM_REQUEST_HEADER } from '../utils/globalSettings';
 
 import '../sass/Event.scss';
 
 const Event = (props) => {
-    const [isLoading, setIsLoading] = useState(false);
-    const [event, setEvent] = useState([]);
-    const [isCheckInReady, setIsCheckInReady] = useState();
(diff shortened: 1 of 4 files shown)


client/src/pages/Event.js
```diff
@@ -1,82 +1,95 @@
...
+            <h4>{event.name}</h4>
+            <p>{moment(event.date).format('dddd, MMMM D, YYYY @ h:mm a')}</p>
+            <p>{event.location.city}</p>
+            <p>{event.location.state}</p>
+            {event.project ? (
+              <Link
+                to={`/project/${event.project._id}`}
+                className="create-form-button"
+                style={{
+                  width: '140px',
```
jbubar, line 69: Screen Shot 2022-06-27 at 5 10 42 PM
not sure why edit event is being cut off here. 

but it looks like we need to add

`height: fit-content;`

in the css

the rest of this looks great

jbubar (changes requested):

client/src/pages/Event.js
FoxShaunR: Pushed a commit that might fix it, but I'm not sure. Which browser are you seeing this on? I don't see it on firefox or chrome/edge

jbubar (approved):
This is so sick!!
Super stoked about it

PR #1158, opened 2022-07-06
Feature/add project leads to project detail page

### LINKED ISSUE #1152: Add a category of "project leads/admins" on the project details page
### Overview
We want to display the leads/admins of a specific project in the details page of that project. It will be a field that the main admin can edit to add/remove leads.

### Action Items
- [x] figure out how to get the project admins (ideally in the back end) 
- [x] display a list of these project admins on the project page by creating an editable field that will hit the managedByUsers endpoint in the projects API
- [x] apply same user validation in the editable field as the rest of the editable fields (only authorized users can edit the field)
- [x] apply the comma separated format for these values

chukalicious:
closes #1152

This PR adds a field for Project Leads in the Project Details page. Only an authorized user can add or remove leads from this field.

### DIFF
--- a/client/src/components/manageProjects/editProject.js
+++ b/client/src/components/manageProjects/editProject.js
@@ -146,6 +146,14 @@ const EditProject = ({
         accessLevel={userAccessLevel}
       />
       <EditableField
+        fieldData={projectToEdit.managedByUsers}
+        fieldName="managedByUsers"
+        updateProject={updateProject}
+        fieldTitle="Managed by Users (comma separated):"
+        accessLevel={userAccessLevel}
+        canEdit={['admin', 'user']}
+      />
+      <EditableField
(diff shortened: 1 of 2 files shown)


client/src/components/manageProjects/editProject.js
```diff
@@ -145,6 +146,14 @@ const EditProject = ({
         fieldTitle="Looking For Description:"
         accessLevel={userAccessLevel}
       />
+      <EditableField
```
FoxShaunR, line 149: I don't think we want to be able to edit this on this page. And I'm pretty sure it wouldn't do anything if we tried to update them using this input. It probably makes more sense to just display this with a "Managed by Users" header and a paragraph element afterwards containing the list of users.

jbubar:
We will probable restart this after we create a new api endpoint for the managedByUsers

PR #1167, opened 2022-07-22
fix: missing body for open and close checkins

### LINKED ISSUE #1165: BUG: The incorrect meetings are display in the checkin page
### Overview
The incorrect meetings are displaying on the checkin page! It is most likely something broken with one of the worker files. The page is displaying the checkin ready meetings that are in the database but the ones that are in the database are not correct. This is really important! Bonnie uses the app in the vrms team meetings

### Action Items
- [x] identify the problem
- [x] Fix it?? 
- [ ] Teach josh how to write better issues - Gonna need a few sprints for this one

### Resources/Instructions
  - which worker file?
    - [backend/workers/createRecurringEvents.js](https://github.com/hackforla/VRMS/blob/development/backend/workers/createRecurringEvents.js)
    - [backend/workers/openCheckins.js](https://github.com/hackforla/VRMS/blob/development/backend/workers/openCheckins.js)
    -  [backend/workers/closeCheckins.js](https://github.com/hackforla/VRMS/blob/development/backend/workers/closeCheckins.js)
<3 thank you!

FoxShaunR:
Closes: https://github.com/hackforla/VRMS/issues/1165

So, we have scheduled jobs to go through and check to see if any recurring meetings are scheduled to begin or end soon. The job fetches all the scheduled events and checks their start times. Then, if they should start or end soon, it calls the update event endpoint for each one that needs to have the checkInReady flag updated to indicate whether or not members can check in to the event.  The only thing the job doesn't do... is actually make any updates to the event record's checkInReady flag.

### DIFF
--- a/backend/workers/closeCheckins.js
+++ b/backend/workers/closeCheckins.js
@@ -50,7 +50,8 @@ module.exports = (cron, fetch) => {
                     headers: {
                       "Content-Type": "application/json",
                       "x-customrequired-header": headerToSend
-                    }
+                    },
+                    body: JSON.stringify({ checkInReady: false })
                 })
                     .catch(err => {
                         console.log(err);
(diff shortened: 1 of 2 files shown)


jbubar (approved):
Looks great

PR #1229, opened 2022-10-17
fix: Update events.js layout and CSS

trillium:
I **believe** these changes don't break anything. I wans't able to test them on my local development server due to me not getting the login working locally.

Rearranged link to be higher up the component tree in event.js so that clicking anywhere on the link goes to the event page

Changed CSS to add border, padding, highlight on hover

### DIFF
--- a/client/src/pages/Events.js
+++ b/client/src/pages/Events.js
@@ -56,11 +56,13 @@ const Events = (props) => {
               <li key={index}>
                 <div key={index} className="list-event-container">
                   <div className="list-event-headers">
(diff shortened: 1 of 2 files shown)


jbubar:
Here are the screen shots for before and after


Screen Shot 2022-10-17 at 7 47 34 PM



Screen Shot 2022-10-17 at 7 46 25 PM

jbubar (approved):
I tested this out, and it is definitely better than the current version. 
I think it will need more work in the future. but it is ready to merge

Issue #1220, opened 2022-10-11
Removal of Fields of Project Information


AmandaGlover-PM:
### Overview
The following fields need to be removed from the front end so nobody can edit. 

- [x] googleDriveId
- [x] lookingDescription
- [x] recruitingCategories
- [x] videoConferenceLink

#### Potential Removal
- [ ] githubIdentifier

JackHaeg:
All of these fields have been removed and are no longer available on the project management screen.

PR #1230, opened 2022-10-18
fix: Comment out fields from issue #1220

trillium:


jbubar (approved):
This is great!
Ill merge it. But we should consider deleting it rather than commenting it out

PR #1247, opened 2022-12-04
(1234) first steps to putting an icon there

evanyang1:


jbubar:
This has been completed in another pr. also we did not want a new library

PR #1342, opened 2023-03-16
Upgrade react scripts (again)

trillium:
Fixes #1324

### What changes did you make and why did you make them ?

This PR was reverted because Eslinst was erroring during development. That *should* be changed now.

#### Key changes:
 - Eslint is properly configured
 - Eslint only provides warnings, doesn't block development

#### 1st Attempt (before revert):
  - Move to react-scripts: ^5.0.1
    - Allows the current applicaton to run on node 18, 19
  - Add babel/core as dev dependency
  - Move images used by the app to client/src/assets to resolve import errors created from changing react-scripts versions.

##### 2nd Attempt:
 - Create .babelrc file at root of monorepo
 - add eslinst-only-warn-package
 - update client/eslinstrc.json
   - Add location of .babelrc
   - Add "only-warn" to plugins
 - delete yarn.lock
 - rebuild yarn.lock

Made this change to close a security vulnerabiltiy and to allow the current application to run on node 18+

### Screenshots of Proposed Changes Of The Website  (if any, please do not screen shot code changes)

<details>
<summary>Visuals before changes are applied</summary>

image

</details>

<details>
<summary>Visuals after changes are applied</summary>
  
image

</details>

### Review Instructions

For anyone that reviews this pull request, in order to test you're going to need to run 

```
cd client
yarn install
nvm use 18
```

in the /client directory

PR #1347, opened 2023-03-19
Revert "Upgrade react scripts (again)"

MattPereira:
Reverts hackforla/VRMS#1342

We need to revert because this PR fails deployment

See action logs
- https://github.com/hackforla/VRMS/actions/runs/4458399332/jobs/7830160345


trillium (approved):
Maybe third time is the charm!

PR #1309, opened 2023-03-05
Fix frontend warnings, minor tweaks

trillium:
Fixes #1302 
Fixes #1303

### What changes did you make and why did you make them ?

  - Removed (boolen ? component : null) pattern, switched to (boolen && component) pattern
  - rename headerToSend on import, save from defining it multiple places for fetch calls
  - Resolve eslint warnings (each commit has notes on resolved warning)

<details>
<summary>Visuals before changes are applied</summary>

image

</details>

<details>
<summary>Visuals after changes are applied</summary>
  
image

</details>

PR #1357, opened 2023-03-22
Fix meeting check in

### LINKED ISSUE #1348: Meeting check-in dropdown showing duplicate options
### Overview
Users of the production version of the VRMS app are currently experiencing a bug when trying to select a meeting to check-in. The drop down list that pops up when users click `--SELECT ONE--` shows the same meeting name repeatedly, which makes it impossible for users to figure out which meeting they should select. 

<details><summary>See screenshot of error reported by user</summary> 

image 

</details>

### Action Items
- [x] Figure out why this bug is occuring
- [x] Refactor the code so it only renders one `<option>` element for each instance of a meeting

### Notes
- The check-in dropdown is only supposed to allow users the option to check into a meeting if that meeting is about to start or currently in progress
- The select input for the meeting check-in dropdown occurs in the `client/src/pages/Home.js` file

### Resources
- Home page component: https://github.com/hackforla/VRMS/blob/development/client/src/pages/Home.js
- User comment from slack: https://hackforla.slack.com/archives/CRGH5HM0Q/p1679093612693469

MattPereira:
Fixes #1348

  - I added the `if(eventExists){ ... } else { ... }` logic to the `filterAndCreateEvents()` function because it was working when this logic existed in the past
      - Trilliums PR from a couple weeks ago removed this logic and it seems to have resulted in many duplicate events being created
           - https://github.com/hackforla/VRMS/pull/1309/files?diff=unified&w=0
  - Ended up just reverting the file to the OG state from before the PR that changed it

### DIFF
--- a/backend/controllers/event.controller.js
+++ b/backend/controllers/event.controller.js
@@ -6,12 +6,13 @@ EventController.event_list = async function (req, res) {
   const { query } = req;
 
   try {
+    console.log('I AM HERE', query);
     const events = await Event.find(query);
+    console.log('EVENTS', events);
     return res.status(200).send(events);
   } catch (err) {
     return res.sendStatus(400);
   }
-
 };
 
 EventController.event_by_id = async function (req, res) {
@@ -45,7 +46,6 @@ EventController.destroy = async function (req, res) {
   } catch (err) {
     return res.sendStatus(400);
   }
-  
 };
 
 EventController.update = async function (req, res) {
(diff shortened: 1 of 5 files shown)


trillium (commented):
Good job figuring this out @MattPereira

trillium (approved):

MicahBear (approved):
Bug squashed. Nice Job, Matt.

PR #1420, opened 2023-07-12
fix: Update openCheckins to use env var

### LINKED ISSUE #1419: Backend using localhost:4000 instead of process.env.BACKEND_PORT
### Overview
Hardcoded port doesn't allow for flexibility for devs to set their own port if they have more than one project running. Switching to BACKEND_PORT from the environment variables solves this.

trillium:
# Was using hardcoded localhost 4000, now uses env var

Fixes #1419

Single line change, change
```js
# from 
const url = process.env.NODE_ENV === 'prod' ? 'https://www.vrms.io' : 'http://localhost:4000';
# to
const url = process.env.NODE_ENV === 'prod' ? 'https://www.vrms.io' : `http://localhost:${process.env.BACKEND_PORT}`;
```
- This change allows devs to designate other ports for their own development environment (I had mine on 4001 and was wondering why it was failing)

### DIFF
--- a/backend/workers/closeCheckins.js
+++ b/backend/workers/closeCheckins.js
@@ -3,7 +3,7 @@ module.exports = (cron, fetch) => {
     // Check to see if any events are about to start,
     // and if so, open their respective check-ins
 
-    const url = process.env.NODE_ENV === 'prod' ? 'https://www.vrms.io' : 'http://localhost:4000';
+    const url = process.env.NODE_ENV === 'prod' ? 'https://www.vrms.io' : `http://localhost:${process.env.BACKEND_PORT}`;
     const headerToSend = process.env.CUSTOM_REQUEST_HEADER;
 
     async function fetchEvents() {
--- a/backend/workers/createRecurringEvents.js
+++ b/backend/workers/createRecurringEvents.js
@@ -8,7 +8,7 @@ module.exports = (cron, fetch) => {
     let RECURRING_EVENTS;
     let TODAY_DATE;
     let TODAY;
-    const URL = process.env.NODE_ENV === 'prod' ? 'https://www.vrms.io' : 'http://localhost:4000';
+    const URL = process.env.NODE_ENV === 'prod' ? 'https://www.vrms.io' : `http://localhost:${process.env.BACKEND_PORT}`;
 
     const headerToSend = process.env.CUSTOM_REQUEST_HEADER;
     const fetchEvents = async () => {
--- a/backend/workers/openCheckins.js
+++ b/backend/workers/openCheckins.js
@@ -3,7 +3,7 @@ module.exports = (cron, fetch) => {
     // Check to see if any events are about to start,
     // and if so, open their respective check-ins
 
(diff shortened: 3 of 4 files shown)


MattPereira:
@Spiteless looks like `localhost:4000` is hardcoded in a few other files as well

Also heads up that only people who start the project by moving into `backend` and `client` and running the start commands individually would be able to take advantage of this solution. Those of us using docker will be still married be married to port 4000 because you have to declare which port to expose and map from inside container to localhost

https://github.com/hackforla/VRMS/blob/224433a9e862ccbebac42ff9dff95791b904e04c/docker-compose.yml#L13-L17

https://github.com/hackforla/VRMS/blob/224433a9e862ccbebac42ff9dff95791b904e04c/backend/workers/createRecurringEvents.js#L11


https://github.com/hackforla/VRMS/blob/224433a9e862ccbebac42ff9dff95791b904e04c/backend/workers/closeCheckins.js#L6


https://github.com/hackforla/VRMS/blob/224433a9e862ccbebac42ff9dff95791b904e04c/backend/workers/slackbot.js#L12

MattPereira (dismissed):
@Spiteless looks like `localhost:4000` is hardcoded in a few other files as well

Also only devs who start the project by moving into `backend` and `client` and running the start commands individually would be able to take advantage of this solution. Those of us using docker will be still married be married to port 4000 because you have to declare which port to expose and map from inside the docker container over to localhost

And if devs aren't using docker they have to go through additional trouble of downgrading their node version from 18 to 16 and they wouldn't have access to `mailhog` on port 4100

https://github.com/hackforla/VRMS/blob/224433a9e862ccbebac42ff9dff95791b904e04c/docker-compose.yml#L13-L17

https://github.com/hackforla/VRMS/blob/224433a9e862ccbebac42ff9dff95791b904e04c/backend/workers/createRecurringEvents.js#L11


https://github.com/hackforla/VRMS/blob/224433a9e862ccbebac42ff9dff95791b904e04c/backend/workers/closeCheckins.js#L6


https://github.com/hackforla/VRMS/blob/224433a9e862ccbebac42ff9dff95791b904e04c/backend/workers/slackbot.js#L12

trillium:
> Also only devs who start the project by moving into backend and client and running the start commands individually would be able to take advantage of this solution. Those of us using docker will be still married be married to port 4000 because you have to declare which port to expose and map from inside the docker container over to localhost
> 
> And if devs aren't using docker they have to go through additional trouble of downgrading their node version from 18 to 16 and they wouldn't have access to mailhog on port 4100

@MattPereira Noted, I'd still like to see the hardcoded instances of localhost shunted to the proper variables. 

There's a couple instances of localhost:3000, and I'm making an assumption that those are there because the backend .env vars don't have a `CLIENT_PORT` variable.
```js
// backend/workers/slackbot.js
...
        const linkNew = 'http://localhost:3000/checkIn/newUser?eventId=' + event._id;
        const linkReturning = 'http://localhost:3000/checkIn/returningUser?eventId=' + event._id;
...
```

Regardless, I still think this would be a useful change -- at least it'll help me specifically 😃

MattPereira (approved):

PR #1448, opened 2023-07-18
feat: Add admin manual CheckInReady toggle

### LINKED ISSUE #1447: Allow Admin to manually toggle if an event is checkInReady
### Overview
- Temp workaround for #1411

We need the ability for an Admin to manually force the events to be check in ready. This allows the team to be on top of potential errors when onboarding is happening and we need the event check in window to be open regardless of what the backend currently says.

### Action Items
 - [ ] Add a button to toggle if an event is Check In Ready

### Limitations
This is a hacky fix, we haven't solved the underlying problem of why service workers might not be handling the `checkInReady` switch as we would expect, but this will lessen the pressure while we figure out a solve.

The service worker that turns off checkin still might catch this and someone will need to manually reopen it (potentially every half hour) until the event is finished.

trillium:
Fixes #1447

  - Create EventsApiService that mimics RecurringApiService
  - Add section that renders events from /events/ that match the project in the projects page
  - Adapt the surrounding components to pass needed props

### Screenshots of Proposed Changes Of The Website 

<details>
<summary>Visuals before changes side by side</summary>

image
image
image
image
image
image
image
image
image
</details>

### DIFF
--- a/client/src/components/manageProjects/editProject.js
+++ b/client/src/components/manageProjects/editProject.js
@@ -17,12 +17,15 @@ const EditProject = ({
   createNewRecurringEvent,
   deleteRecurringEvent,
   updateRecurringEvent,
+  regularEvents,
+  updateRegularEvent
 }) => {
   // Add commas to arrays for display
   const partnerDataFormatted = projectToEdit.partners.join(', ');
   // eslint-disable-next-line no-unused-vars
   const recrutingDataFormatted = projectToEdit.recruitingCategories.join(', ');
-  const [rEvents, setREvents] = useState([]);
+  const [recurringEventsState, setRecurringEventsState] = useState([]);
+  const [regularEventsState, setRegularEventsState] = useState([]);
   const [selectedEvent, setSelectedEvent] = useState();
   const [isCreateNew, setIsCreateNew] = useState();
 
@@ -31,16 +34,28 @@ const EditProject = ({
 
   // Get project recurring events when component loads
   useEffect(() => {
+    if (regularEvents) {
+      setRegularEventsState(
+        regularEvents
+          // eslint-disable-next-line no-underscore-dangle
+          .filter((e) => e?.project?._id === projectToEdit._id)
+          .map((item) => ({...item, ...readableEvent(item), raw: item}))
+          .reverse() // sorts most recent events first
+      );
+    }
+  }, [projectToEdit, regularEvents, setRegularEventsState]);
+
+  useEffect(() => {
     if (recurringEvents) {
-      setREvents(
+      setRecurringEventsState(
         recurringEvents
           // eslint-disable-next-line no-underscore-dangle
           .filter((e) => e?.project?._id === projectToEdit._id)
-          .map((item) => readableEvent(item))
+          .map((item) => ({...item, ...readableEvent(item)}))
           .sort((a, b) => a.dayOfTheWeekNumber - b.dayOfTheWeekNumber)
-      );
-    }
-  }, [projectToEdit, recurringEvents, setREvents]);
+          );
+        }
+  }, [projectToEdit, recurringEvents, setRecurringEventsState]);
 
   return (
     <div>
@@ -179,7 +194,7 @@ const EditProject = ({
         <h3>Recurring Events</h3>
         <h2 className="event-alert">{eventAlert}</h2>
         <ul>
-          {rEvents.map((event) => (
+          {recurringEventsState.map((event) => (
             // eslint-disable-next-line no-underscore-dangle
             <li key={`${event.event_id}`}>
               <button type="button" onClick={() => setSelectedEvent(event)}>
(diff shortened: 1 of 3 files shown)


client/src/components/manageProjects/editProject.js
```diff
@@ -36,7 +36,7 @@ const EditProject = ({
         recurringEvents
           // eslint-disable-next-line no-underscore-dangle
           .filter((e) => e?.project?._id === projectToEdit._id)
-          .map((item) => readableEvent(item))
+          .map((item) => ({...item, ...readableEvent(item)}))
```
trillium, line 54: This change is allows `checkInReady` to be passed via props

client/src/components/manageProjects/editProject.js
```diff
@@ -207,4 +198,21 @@ const EditProject = ({
...
+
+  return (
+    <li key={`${event.event_id}`}>
+      <button type="button" onClick={() => {console.log(event);setSelectedEvent(event)}}>
+        <div>{event.name}</div>
+        <div className="event-list-details">
+          {`${event.dayOfTheWeek}, ${event.startTime} - ${event.endTime}; ${event.eventType}`}
+          <div className="edit-icon"><EditIcon /></div>
+        </div>
+        <div className="event-list-description">{`Can uses check into this event now?: ${event.checkInReady ? "Yes" : "No"}`}</div>
```
Spiteless, line 211: `setSelectedEvent` designates which event should be editable in the modal rendered by `<EditMeetingTimes />`

client/src/components/manageProjects/editProject.js
```diff
@@ -207,4 +198,21 @@ const EditProject = ({
   );
 };
 
+function RecurringEvent({event, setSelectedEvent}) {
+
+  return (
+    <li key={`${event.event_id}`}>
+      <button type="button" onClick={() => {console.log(event);setSelectedEvent(event)}}>
```
trillium: `setSelectedEvent` designates which event should be editable in the modal rendered by `<EditMeetingTimes />` and passes on all the `event` values.

client/src/components/manageProjects/editableMeeting.js
```diff
@@ -15,6 +15,7 @@ const EditableMeeting = ({
   handleEventDelete,
   formErrors,
   videoConferenceLink = '',
+  checkInReady,
```
trillium: Ensures `checkInReady` is destructed off of props

client/src/components/manageProjects/editableMeeting.js
```diff
@@ -76,6 +77,13 @@ const EditableMeeting = ({
         >
           DELETE
         </button>
+        <button
+          type="button"
+          className="create-form-button"
+          onClick={handleEventUpdate(eventId, {checkInReady: !checkInReady})}
+        >
+          ToggleCheckIn({!checkInReady ? "Yes" : "No"})
+        </button>
```
trillium: `handleEventUpdate()` takes an `evendId`, a `values` object, and two other props unused here.

The `values` object is passed an obj with the opposite of `checkInReady`, and sent to the db updater fuction.

client/src/components/manageProjects/editMeetingTimes.js
```diff
@@ -106,6 +106,13 @@ const EditMeetingTimes = ({
         };
       }
 
+      if ("checkInReady" in values) {
+        theUpdatedEvent = {
+          ...theUpdatedEvent,
+          checkInReady: values.checkInReady,
+        };
+      }
+
```
trillium: This section updates `handleEventUpdate()` in the same style as the original function, allowing minimal change for us to update `checkInReady` in the db.

MattPereira:
@Spiteless your changes do successfully update the associated `recurringEvent` object in the db, however, it does't show up on the check in page for users who are not logged in because the home page component is making a request to the `events` endpoint. For anyone who is unfamiliar, the `recurringEvents` are used to create `events` via the [createRecurringEvents.js](https://github.com/hackforla/VRMS/blob/224433a9e862ccbebac42ff9dff95791b904e04c/backend/workers/createRecurringEvents.js) worker. So there is theoretically an `event` object in the db for every single time a meeting happens which results in the 16,095 `events` records that are in the prod db right now, but there are only 67 `recurringEvents` which generate all those `events`

I refactored the home page a few months ago and left it hitting `api/events` because that's how it was before and didn't want to break anything but maybe swapping it out for `api/recurringEvents` could be an option to explore? 

One thing to keep in mind is that if you manually flip it to `checkInReady: true` you would also have to manually flip it back to `checkInReady: false` or the meeting would remain as always check in ready. If you swap the endpoint to `/api/recurringEvents?checkInReady=true` on your local env you will see about 40 `recurringEvents` sporting a check in ready true coming from the test db 

https://github.com/hackforla/VRMS/blob/224433a9e862ccbebac42ff9dff95791b904e04c/client/src/pages/Home.js#L21

MattPereira:
> maybe swapping it out for api/recurringEvents could be an option to explore?

This might not be a good idea since the workers only automate the open and close of `events` so changing the endpoint that gets hit by the homepage to `/api/recurringEvents?checkInReady=true` would also require changing the worker files that open and close the `events` to target `recurringEvents` instead

client/src/components/manageProjects/editProject.js
```diff
@@ -207,4 +198,21 @@ const EditProject = ({
   );
 };
 
+function RecurringEvent({event, setSelectedEvent}) {
+
```
evanyang1: Small nit: remove empty line

evanyang1 (approved):
LGTM! Other perspectives welcome

MattPereira:
@evanyang1 how did you test that users would be able to check into meetings that are manually flipped? 

am i misunderstanding the goal for this PR?

Screenshot from 2023-07-19 17-46-09

client/src/components/manageProjects/editProject.js
bkmorgan3: check spelling

trillium:
Updated this commit

MattPereira (approved):
@Spiteless hey man well done figuring this one out! :beers: 

The feature works as expected, but in the process of testing I discovered that duplicate events are sneaking into the test db

<details><summary>See screenshot</summary>

Screenshot from 2023-07-23 15-16-53

</details>


Luckily, it doesn't appear as though there are duplicate events in the prod db

<details><summary>See screenshot</summary>

Screenshot from 2023-07-23 15-29-34

</details>


Therefore, I believe your PR will successfully provide PMs with the ability to manually flip the event to `checkInReady: true`

PR #1454, opened 2023-07-28
fix: Add query to return events within 3 weeks

### LINKED ISSUE #1453: checkInReady Toggle pulls too many events from the db, is  real slow
### Overview
Need to clamp down then amount of enents returned to make the query faster.

trillium:
Fixes #1453

The PR I made to force checkInReady pulls all the events from the db, taking a long time and leading the app to stall. It will eventually return the query populated with data but it's too slow to fit the purpose.

Adding a query string that mongodb recognizes should fix the issue.

  - Add query string to return only events that started within the past 3 weeks

### Screenshots of Proposed Changes Of The Website.
<details>
<summary>Visuals side by side</summary>

 * RIght is previous, left is with changes, returns much less events.
image

</details>

### DIFF
--- a/client/src/api/EventsApiService.js
+++ b/client/src/api/EventsApiService.js
@@ -11,7 +11,13 @@ class EventsApiService {
 
   async fetchEvents() {
     try {
-      const res = await fetch(this.baseUrl, {
+      const threeWeeksInMilliseconds = 3 * 7 * 24 * 60 * 60 * 1000; // 3 weeks in milliseconds
+      const threeWeeksAgo = new Date()
+      threeWeeksAgo.setTime(threeWeeksAgo.getTime() - threeWeeksInMilliseconds)
+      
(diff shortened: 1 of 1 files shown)


client/src/api/EventsApiService.js
```diff
@@ -11,7 +11,13 @@ class EventsApiService {
 
   async fetchEvents() {
     try {
-      const res = await fetch(this.baseUrl, {
+      const threeWeeksInMilliseconds = 3 * 7 * 24 * 60 * 60 * 1000; // 3 weeks in milliseconds
+      const threeWeeksAgo = new Date()
```
evanyang1: Recommended: semicolons

trillium: Added, please review again

trillium:
IMPORTANT:

Bonnie asks that this feature only apply to onboarding. Okay to merge now, but will still need to be updated

jbubar (approved):
I def tested it and it def works and Im good at my job

PR #1455, opened 2023-07-30
Implement React Hook Form attrs

### LINKED ISSUE #1387: Add New Project Form Validation - Location
Under Location, we should add validation to this field so that it requires a string that contains "zoom.us"

     
AC1:
GIVEN that a user adds a new project
WHEN they enter a url that contains the string "zoom.us" under location
THEN the field will not reject the entry
AND allow the user to save as long as all other required fields are entered

AC2:
GIVEN that a user adds a new project
WHEN they enter a url that does not contain the string "zoom.us" under location
THEN text box will turn red
AND error message underneath box will read "please enter valid zoom link"
AND user will not be able to save

bkmorgan3:
Fixes #1387 

  - Implement React Hook Form attributes so it can control form inputs and check for validation errors
  - added regex check on Location inside SimpleInputs 
  -  added non-null validation on each input since each input is rendered via a Map call
  - displayed validation errors in red at the bottom of each input
  - Clean up ProjectAPI-Create call, removing unused properties and streamline req.body

<details>
  
Screenshot 2023-07-30 at 3 56 30 PM
Each Input has basic non-null validation

Screenshot 2023-07-30 at 3 32 35 PM

</details>


client/src/components/ProjectForm.js
```diff
@@ -235,36 +210,14 @@ export default function ProjectForm() {
...
-                    input.name === 'location'
-                      ? locationType === 'remote'
-                        ? 'Enter project zoom link'
-                        : 'Enter project street address'
-                      : input.placeholder
-                  }
-                  variant="outlined"
-                  type={input.type}
-                  onChange={handleChange}
-                  helperText=" "
```
trillium, line 255: Would be great to utilize MUI's helperText prop for TextArea here:

```js
<TextField
  helperText={errors[input.name]?.message || " "} //  This will keep spacing to prevent layout shift
  error={errors[input.name]?.message ? true : false}
  ...
/>

File: client/src/components/ProjectForm.js [hunk 2]
```diff
@@ -235,36 +210,14 @@ export default function ProjectForm() {
...
-                        px: '4px',
-                        py: '5px',
-                      },
-                    },
-                  })}
+                 type={input.type}
+                  {...register(input.name,  {required: `${input.name} is required` , pattern: {value: input.value, message: `${input.errorMessage}`}})}
+                 placeholder={input.placeholder}
                 />
+                <p style={{ color: '#fa114f' }}>{errors[input.name]?.message}</p>
```
trillium: Please drop this change and use the `helperText` prop from the above comment in line 255

client/src/components/ProjectForm.js
```diff
@@ -235,36 +210,14 @@ export default function ProjectForm() {
                       {input.label}
                     </InputLabel>
                   </Grid>
-
                   {input.name === 'location' && locationRadios}
                 </Grid>
-
                 <TextField
-                  id={input.name}
```
trillium, line 243: It would be great to validate the form on blur when the user navigates away from the form. Right now the form only validates the data on Submit, I think there's a way to have the form validate the data onBlur, but if there isn't an obvious way via  config object you can add the `onBlur` prop with a callback function.

```js
<TextArea
  onBlur={() => trigger(input.name)}
  ... 
/>

trillium (changes requested):
Looking good, I like where these changes are going.

Requesting changes to
 * the location of the error text is displayed (move `<p>` to `helperText` prop - info in comments). 
 * validate form onBlur

Thanks for your hard work @bkmorgan3  :)

client/src/components/ProjectForm.js
bkmorgan3: @Spiteless  fixed.

client/src/components/ProjectForm.js
bkmorgan3: @Spiteless   fixed

trillium:
Notes from todays call:

@Spiteless
* Issue with rendering location (in-person) and location (remote)
* Advise making separate component that handles edge cases via props
* Still want to use current style of simpleInputs
* Want to be able to add validations to the simple input object
  * Can we add validations in an array that conform with react-hook-form validator object? [This page - search validate functions](https://www.react-hook-form.com/api/useform/register)

@bkmorgan3 mentioned blocker via slack on how to render location object. @Spiteless will play around with this to see if we can figure out how to render and validate that between the different location field types.

bkmorgan3:
@Spiteless 
please review again.

trillium (approved):
Thanks Brad, happy to merge this PR :)

PR #1499, opened 2023-09-04
trim values before accessing DB

### LINKED ISSUE #1402: Trim trailing spaces of all fields
### Overview
We want to sanitize user inputs so that they do not accidentally add a trailing space to the end of some data like project name.

### Action Items
- [ ] Trim trailing spaces from project input fields on the front end for project create and update

### Resources/Instructions
Please see https://github.com/hackforla/website/issues/4874#issuecomment-1606155521

bkmorgan3:
Fixes #1402 

  - On the Project Model for each input that correlates to a Project Field on the add / edit form I added a trim property to remove extra white space on both ends of the value.
  - Added to the Model because if a user bypasses the UI through Postman or something, the values are still trimmed and it places the burden of doing the work on the DB - as light as an operation as it is.
  -

No Screenshots but you confirm this works by Adding and Editing a project with spaces and checking the project in Mongo Compass.

### DIFF
--- a/backend/models/project.model.js
+++ b/backend/models/project.model.js
@@ -16,19 +16,19 @@ Idea for the future: programmingLanguages, numberGithubContributions (pull these
 */
 
 const projectSchema = mongoose.Schema({
-    name: { type: String },
-    description: { type: String },
-    githubIdentifier: { type: String },
+    name: { type: String, trim: true },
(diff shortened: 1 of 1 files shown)


trillium (approved):
Works, thanks @bkmorgan3 !

PR #1539, opened 2023-11-05
Ts.1411

### LINKED ISSUE #1411: Bug: Users Cannot Check In
### Dependency
- [ ] Close if the following issue is completed: https://github.com/hackforla/VRMS/issues/1872

### Overview
Currently, users cannot check in to VRMS when visiting www.vrms.io. There is no check in button available for a user to select and check in to VRMS. 

Desired behavior for the meeting check-in is to allow people to check in to a meeting 30 min before starts and 30 min after the meeting ends

### Action Items
- [x] Check to see what the current behavior is. Look into events created 
- [x] Move the following issue from Icebox to In Progress and self assign. Use it to check if Guides team check in is working properly and communicate with stakeholder after fix: https://github.com/hackforla/VRMS/issues/1530

### Resources/Instructions
- This issue is part of the epic: https://github.com/hackforla/VRMS/issues/1838
- See attached screenshot of existing VRMS landing page. 

<details><summary>Screenshot from www.vrms.io</summary>
<p>
Screenshot 2023-07-03 at 10 18 29 PM
</p>
</details> 

<details><summary>Screenshot from dev.vrms.io</summary>
<p>
Screenshot 2023-07-03 at 10 28 24 PM
</p>
</details> 

- This issue is part of the epic #1838

trillium:
Fixes #1411

  - Shift code from if/then block to helper function
  - Switch event.date to event.startTime

The bug we were experiencing was because events were being based off of the `event.date` property rather than the `event.startTime` property. When the `createAndFilterEvents` worker ran, it was using old data that the user wasn't setting within the app.

There are already helper functions made on the front end to transfer a timestamp from `6:00 PM` to a valid Date object. The recurring events model has these saved in theory at the right spots, but it is also recommended that project managers update their start times of their events within the app.

If they wish to keep their start times, please update them twice eg `7:00 PM` to `6:00 PM` and back to `7:00 PM`. This way we can ensure that the correct startTime is saved in the `db`.

### DIFF
--- a/backend/workers/createRecurringEvents.js
+++ b/backend/workers/createRecurringEvents.js
@@ -60,50 +60,56 @@ module.exports = (cron, fetch) => {
             // forEach function with async/await.
             for (filteredEvent of filteredEvents) {
                 const eventExists = await checkIfEventExists(filteredEvent.name);
-                // console.log('Event exists? ', eventExists);
-                const eventDate = new Date(filteredEvent.date);
 
                 if (eventExists) {
+                    //Do nothing
                     console.log("Not going to run ceateEvent");
                 } else {
                     // Create new event
-                    const hours = eventDate.getHours();
-                    const minutes = eventDate.getMinutes();
-                    const seconds = eventDate.getSeconds();
-                    const milliseconds = eventDate.getMilliseconds();
-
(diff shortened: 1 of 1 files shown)


bkmorgan3 (approved):
yolo -- per Trillium

PR #1541, opened 2023-11-10
fix: Handle event.date being undefined in closeCheckins

### LINKED ISSUE #1540: Handle null/undefined in closeCheckins for event.date prop
### Overview
Handle case where a db object doesn't have `event.date` in the events object. This resolves an error where closeCheckins fails before completing its run, allowing events to stay open beyond their allotted time.

### Action Items
- [ ] Handle no `event.date` prop

REPLACE THIS TEXT - If the issue has already been researched, and the course of action is clear, this will describe the steps.  However, if the steps can be divided into tasks for more than one person, we recommend dividing it up into separate issues or assigning it as a pair programming task.

### Resources/Instructions
REPLACE THIS TEXT - If there is a website that has documentation that helps with this issue provide the link(s) here.

trillium:
Fixes #1540

  - Add `if` block to check if `.date` prop exists on events object

### DIFF
--- a/backend/workers/closeCheckins.js
+++ b/backend/workers/closeCheckins.js
@@ -28,6 +28,12 @@ module.exports = (cron, fetch) => {
         if (events && events.length > 0) {
 
             const sortedEvents = events.filter(event => {
+                if (!event.date) {
+                    // handle if event date is null/undefined
+                    // false meaning don't include in sortedEvents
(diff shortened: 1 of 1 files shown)


jbubar (approved):
LGTM. I really hope this works and fixes the events bug.

PR #1627, opened 2024-04-16
New fetch logic

### LINKED ISSUE #1619: Security leak: Rewrite the ProjectList.js page to only call necessary project data when pm's are viewing
### Overview
Currently on the Project List page it is calling all of the project data no matter what the user access level is. This means that a user can grab any project data and do a post request. This is bad practice and a security risk. It would be better and more secure if when a project manager is viewing we only call for the projects that they have access to. 

### Action Items
- [x] Log in with PM account that is associated with a single project ([see comment](https://github.com/hackforla/VRMS/issues/1619#issuecomment-2019369543) for steps to create a PM account via an admin account)
- [x] Navigate to Projects page
- [x] view client/src/pages/ProjectList.js
- [x] If the user is a PM, only do a get request for the projects they are assigned to
- [x] test to see if it still works
- [x] make a pr

### Resources/Instructions
Admin users can
 *  - see all projects in database
 *  - see button to add a new project
 
Project managers can
 *   - see all projects they manage
 *   - will not see button to add a new project
 
 
 
image

bkmorgan3:
Fixes #1619

This PR might open up a few further questions. 
First, Can we guarantee that a User will only PM one project?  I hardcoded a fetch based off only 1 ID in a users Managed Projects array.  That would introduce a slippery slope of additional requests if there are more than one.  if, 2, if 3, etc.  

  - I added a check for accessLevel and fetch project info accordingly
  - render the name of the project(s)
  

<details>

Screenshot 2024-04-16 at 3 28 05 PM

</details>

<details>
  
Screenshot 2024-04-16 at 2 00 58 PM

</details>


JackHaeg:
**Per Trillium - Changes requested:** 
- [ ] Drop yarn lock changes
- [x] Configure fetch call to accept multiple projects (U
  - filter the list server side
  - OR Add query perimeters to the MongoDB call

bkmorgan3:
@Spiteless  I opted for filtering on server because I didn't find a way to use multiple calls to Mongo. 
For testing a PM for 2 ids I used  this 
user.managedProjects.push("6453dfff390bc1002ab3eb6d")
inside the check where user?.accessLevel !== 'admin, line 53ish  in ProjectList.

client/src/pages/ProjectList.js
```diff
@@ -70,7 +67,7 @@ export default function ProjectList() {
   }
 
   // Render loading circle until project data is served from API
-  if (!projects)
+  if (!project && !projects)
```
trillium, line 70: This is the same test twice, was it meant to be something else?

trillium: Nevermind, I see that its project singular, and projects, multiple. My bad

trillium: It would help to change the names of these variables so they're more clear though

client/src/pages/ProjectList.js
```diff
@@ -99,7 +96,21 @@ export default function ProjectList() {
       )}
 
       <TitledBox title="Active Projects" childrenBoxSx={{ p: 2 }}>
-          {projects.map((project) => (
```
trillium, line 102: This looks like it can be refactorered so that we're not duplicating the component structure.

How do you feel about defining a `projectName`, `projectTo`, `projectId` and maybe `projectsList` properties in the function body based on  the differences in then `user?.accessLevel` map so we don't need to redefine the component structure?

If you feel like this is more clear as written that's fine too

yarn.lock
trillium, line 1: Can you restore this file to an unchanged state

backend/routers/projects.router.js
```diff
@@ -7,6 +7,8 @@ const { AuthUtil } = require("../middleware");
 // The base is /api/projects
 router.get('/', ProjectController.project_list);
 
+router.put('/', ProjectController.projects)
```
trillium: This feels like we don't want this to a `PUT` request because the client is only requesting information, not updating anything. If we were to turn this into a `GET` request what would need to change in order for it to work properly?

trillium (commented):
Added a few comments, leme know what you think!

bkmorgan3:
My rebase / merge game is not tight

PR #1644, opened 2024-05-29
refactor project filtering for PMs

### LINKED ISSUE #1619: Security leak: Rewrite the ProjectList.js page to only call necessary project data when pm's are viewing

bkmorgan3:
Fixes #1619 

  - refactor logic for fetching projects based on role
  - filter PM projects on the server
  
For Testing you will need an Admin account to fetch all projects and a non-admin PM account with one project and if needed you can push a project into user.managedProjects to see how it works with more than 1 project.

### DIFF
--- a/backend/routers/projects.router.js
+++ b/backend/routers/projects.router.js
@@ -7,6 +7,8 @@ const { AuthUtil } = require("../middleware");
 // The base is /api/projects
 router.get('/', ProjectController.project_list);
 
+router.put('/', ProjectController.projects);
+
 router.post('/', AuthUtil.verifyCookie, ProjectController.create);
 
 router.get('/:ProjectId', ProjectController.project_by_id);
--- a/backend/routers/projects.router.js
+++ b/backend/routers/projects.router.js
@@ -7,7 +7,8 @@ const { AuthUtil } = require("../middleware");
 // The base is /api/projects
 router.get('/', ProjectController.project_list);
 
-router.put('/', ProjectController.projects);
+// Its a put because we have to send the PM projects to be filtered here
+router.put('/', ProjectController.pm_filtered_projects);
 
 router.post('/', AuthUtil.verifyCookie, ProjectController.create);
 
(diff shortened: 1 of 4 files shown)


backend/routers/projects.router.js
```diff
@@ -7,6 +7,8 @@ const { AuthUtil } = require("../middleware");
 // The base is /api/projects
 router.get('/', ProjectController.project_list);
 
+router.put('/', ProjectController.projects);
```
trillium: Can we refactor this name? Maybe `project_list_pm_filtered`

trillium: Can we add an explanation comment that this request is retruning a PMs project list.

client/src/pages/ProjectList.js
```diff
@@ -42,21 +42,18 @@ export default function ProjectList() {
   useEffect(
     function getProjectsOnMount() {
       async function fetchAllProjects() {
```
trillium, line 44: Nice job, I really like this refactor

trillium (changes requested):
This is great, thanks for putting the work in to understand and resolve this.

I've requested a method name change for ProjectController.projects, and a comment in the projects.router.js to explain the route so people who have less context later on can get on the trail of how this works faster.

Thanks Brad!

trillium (approved):
Great work !

PR #1768, opened 2024-10-01
feat: Refresh yarn.lock, add @swc/core

trillium:
Test branch to investigate if adding @swc/core fixes front end build action


trillium:
YOLO merge

PR #1799, opened 2024-10-21
1771 - API for Fetching Project Managers 

### LINKED ISSUE #1771: Create API call to get all users who are listed as managers on a project
### Overview
Currently, there is no way for admins to see a list of users who are project managers and which projects they are associated with on VRMS. Given that, we need to create an API that will return a list of users with each entry corresponding to a project they manage.  

### Action Items
- [x] Reference best practices research on #1769
- [x] Create MongoDB query that returns a list of user objects where the user has a list of projects they manage. Only return users who manage projects.
- [x] Ensure that the user objects are populated with the project name and project ID of what they manage.

### Resources/Instructions
- Refer to: https://www.mongodb.com/docs/
- This issue is part of this epic: #1737

ntrehan:
Fixes #1771

  - Added an API that fetches all the project Managers and the projects they manage

<details>
<summary>The API request and response</summary>
image

</details>

### DIFF
--- a/backend/controllers/project.controller.js
+++ b/backend/controllers/project.controller.js
@@ -67,4 +67,26 @@ ProjectController.destroy = async function (req, res) {
 };
 
 
+ProjectController.getProjectManagers = async function (req, res) {
+  try {
+    const userProjectMap = {};
+    const projects = await Project.find({
+      managedByUsers: { $exists: true, $ne: [] }
+    });
+
+    for (const project of projects) {
+      for (const user of project.managedByUsers) {
+        if (!userProjectMap[user]) {
+          userProjectMap[user] = [];
+        }
+        userProjectMap[user].push(project.name);
+      }
+    }
+    return res.status(200).send(userProjectMap);
+  } catch (err) {
+    return res.sendStatus(400);
+  }
+};
+
+
 module.exports = ProjectController;
--- a/backend/routers/projects.router.js
+++ b/backend/routers/projects.router.js
@@ -5,6 +5,9 @@ const { ProjectController } = require('../controllers');
 const { AuthUtil } = require("../middleware");
 
 // The base is /api/projects
+
+router.get('/projectManagers', ProjectController.getProjectManagers);
+
 router.get('/', ProjectController.project_list);
 
 // Its a put because we have to send the PM projects to be filtered here


pluto-bell:
@ntrehan just a heads up that your API req & res screenshot is currently broken / doesn't populate.

ntrehan:
> @ntrehan just a heads up that your API req & res screenshot is currently broken / doesn't populate.

Hi @pluto-bell  Thanks for pointing that out!
However no matter how many times I reupload it there
I am not able to get it fixed 
image
So I'll just add it here

backend/controllers/project.controller.js
```diff
@@ -67,4 +67,26 @@ ProjectController.destroy = async function (req, res) {
...
+    const projects = await Project.find({
+      managedByUsers: { $exists: true, $ne: [] }
+    });
+
+    for (const project of projects) {
+      for (const user of project.managedByUsers) {
+        if (!userProjectMap[user]) {
+          userProjectMap[user] = [];
+        }
+        userProjectMap[user].push(project.name);
```
jng34, line 82: Ref: #1771 
@ntrehan It seems like you might have missed the capture of the `project.id`. Perhaps pushing an project object with `id` and `name` properties would be great. Other than that, the code block looks good!

vorleakyek:
If `curl.exe --location 'http://localhost:3000/api/projects/projectManagers'` returns "Unauthorized", 
try  
curl.exe --location 'http://localhost:3000/api/projects/projectManagers' --header 'x-customrequired-header: custom-header-here' 
Note: replace custom-header-here with your own header. 

Note: it returns the test data because we're running it locally. It will the actual data for the production environment. 

`{
  : [
    "VRMS"
  ],
  lead one: [
    "New Test Project A+"
  ],
  as: [
    "Fancy New Project"
  ],
  qegq: [
    "! ! ! ! ! ! ! ! ! 🚨! ! ! DOGEfdsafdasfdsa"
  ],
  test: [
    "! ! ! ! !TESTING123fdsaf"
  ],
  user: [
    "!  ! ! ! ! ! ! Project !!!! - testing 🐢fdsaffds"
  ],
  a: [
    "TESTING Editing",
    "Projj"
  ],
  none: [
    "meep"
  ]
}`

backend/controllers/project.controller.js
trillium: This is a good catch

backend/controllers/project.controller.js
```diff
@@ -67,4 +67,26 @@ ProjectController.destroy = async function (req, res) {
 };
 
 
+ProjectController.getProjectManagers = async function (req, res) {
+  try {
+    const userProjectMap = {};
```
trillium, line 72: I think we want an array of objects of shape:

``` js
const userProjectMap = []; //changed to array
```

```js
// Object shape from API call
{
    user, // will this just be the userId string or will
          // we have a populated user object with email, etc?
    project: {
        name: string
        projectId: string
    }
}
```

Does that look right to you? Then we'd be returning an array that can be sorted more easily on the front end with some sort functions

backend/controllers/project.controller.js
```diff
@@ -67,4 +67,26 @@ ProjectController.destroy = async function (req, res) {
...
+
+    for (const project of projects) {
+      for (const user of project.managedByUsers) {
+        if (!userProjectMap[user]) {
+          userProjectMap[user] = [];
+        }
+        userProjectMap[user].push(project.name);
+      }
+    }
+    return res.status(200).send(userProjectMap);
```
trillium, line 85: Am I right in understanding that the final out of the array (or object if thats the structure we go with) will only have user IDs? Do we need to make a second db call to populate the user object?

trillium (changes requested):
Requesting clarification and changes on API

 - will user object be populated for each entry?
 - Can we switch to an array instead of an object so that the frontend doesnt need to duplicate data for users with more than one managed project?
 
 Thanks for putting so much work into getting this over then hill!

trillium:
PR has been added to [user_permissions_search](https://github.com/hackforla/VRMS/tree/user_permissions_search) feature branch!

PR #1902, opened 2025-04-02
1900 reconfig backend tests

### LINKED ISSUE #1900: Reconfigure Jest for new backend integration tests and clear outdated configs
### Overview
We are updating our integration tests such that they should not reset the state of the test databases between tests. We need to reconfigure jest and our set-up modules to support this. Additionally, there are some old and unused configurations that should be removed.

### Action Items
- [x] Remove configs related to the `@shelf/jest-mongodb` preset
- [ ] Write a new method in `setup-test.js` that sets up a `MongoMemoryServer` but doesn't `removeAllCollection` `afterEach`

### Resources/Instructions 
- This issue is part of the epic: #1899

dannyprikaz:
Fixes #1900 

  - Removed any old configs related to `@shelf/jest-mongodb`
  - Created a new method in `setup-test.js` for creating a database without clearing the collections after each test
  - Changed the test environment to `node` because the default (`jsdom`) makes Mongoose log a warning on each test


trillium (approved):
Thanks so much for your work Danny, we reviewed it in the call

Issue #1713, opened 2024-07-30
Update HTML components to MUI: ./src/pages/ProjectLeaderDashboard.js


JackHaeg:
### Overview
Review all components in`./src/pages/ProjectLeaderDashboard.js` and replace all standard HTML components with applicable MUI components.

### Action Items
- [ ] Review components in:`./src/pages/ProjectLeaderDashboard.js` and search for any standard HTML components.
  - _IF all components are already updated to relevant MUI_
    - [ ] Comment on this issue that all components are already updated to MUI. 
    - [ ] Close the issue & check the box associated with this issue on the [MUI issue epic](https://github.com/hackforla/VRMS/issues/1617)
    - [ ] Move on to the next [MUI issue](https://github.com/hackforla/VRMS/issues/1617) 
  - _IF standard HTML components are found in the file:_ 
    - [ ] Replace these standard HTML components with applicable MUI components. 
    - [ ] Make a PR for this file

### Resources
- https://mui.com/material-ui/
- If you run into any questions, add a comment to the issue and tag "@trillium"

jng34:
@JackHaeg @trillium I have replaced the HTML components with MUI components. How do I go about checking to see if the components are rendered correctly on the page? Right now, the endpoint `/projectleader`, which renders the `ProjectLeaderDashboard` page, is empty. 

<img width="610" alt="Image" src="https://github.com/user-attachments/assets/d26552a2-887e-425c-9a13-11fe69f14243" />

JackHaeg:
@trillium Do you have pointers for James on this?

JackHaeg:
@trillium when you have a moment, can you please provide feedback to James on this comment above: https://github.com/hackforla/VRMS/issues/1713#issuecomment-2654780332 ?

JackHaeg:
Discussed with @trillium, we are closing this issue as unplanned, because this is an unused component of VRMS at the moment.

PR #1906, opened 2025-04-22
Unit Testing for CheckIns Router

### LINKED ISSUE #1886: Create New Unit Tests for ./backend/routers/checkIns.router.js
### Overview
Create unit tests for each route on `./backend/routers/checkIns.router.js`

### Action Items
- [x] Create tests for every route described in `./backend/routers/checkIns.router.js`, isolating their functionality from all dependencies

### Resources/Instructions
Example: https://github.com/hackforla/VRMS/pull/1882/files
- This issue is part of the epic: https://github.com/hackforla/VRMS/issues/1880

jng34:
Fixes #1886 

  - Wrote unit tests for `checkIns.router.js` that isolate the behavior of the router from its dependencies

### Screenshot of passed tests
image


backend/routers/checkIns.router.test.js
```diff
@@ -0,0 +1,116 @@
...
+        id: 'user2',
+        name: {
+          firstName: 'test',
+          lastName: 'user',
+        },
+      };
+
+      // Mock Mongoose methods
+      User.findById.mockResolvedValue(mockUser);
+      CheckIn.find.mockReturnValue({
```
dannyprikaz, line 66: Unused

backend/routers/checkIns.router.test.js
```diff
@@ -0,0 +1,116 @@
...
+        populate: jest.fn().mockResolvedValue({ ...mockCheckIns[1], userId: 'user2' }),
```
dannyprikaz: This function can just return `mockCheckins[1]` and everything will work fine. Or if you want to retun `mockCheckins` that would work too as long as you change line 85 as well.

backend/routers/checkIns.router.test.js
```diff
@@ -0,0 +1,116 @@
+// Mock and import CheckIn model
+jest.mock('../models/checkIn.model');
+jest.mock('../models/user.model');
+const { CheckIn, User } = require('../models');
```
dannyprikaz: Don't need User

backend/routers/checkIns.router.test.js
```diff
@@ -0,0 +1,116 @@
...
+
+// Import the check-ins router
+const express = require('express');
+const supertest = require('supertest');
+const checkInsRouter = require('./checkIns.router');
+
+// Create a new Express application for testing
+const testapp = express();
```
dannyprikaz, line 11: add a line `testapp.use(express.json());`

This configures body parser middleware, such that the req.body of any post methods won't be `undefined`.

This fixes your ability to check that `Checkin.create` is called with what you expect it to be called with.

backend/routers/checkIns.router.test.js
```diff
@@ -0,0 +1,116 @@
...
+        createdDate: String(new Date()),
+      };
+
+      // Mock create method
+      CheckIn.create.mockResolvedValue(newCheckIn);
+
+      const response = await request.post('/api/checkins').send(newCheckIn);
+
+      // Tests
+      expect(CheckIn.create).toHaveBeenCalled();
```
dannyprikaz: Once you add the configuration for body parser middleware you can do `expect(CheckIn.create).toHaveBeenCalledWith(newCheckin);

dannyprikaz (changes requested):
Hey, @jng34 

Great work on this. Nice job in particular with mocking the `populate` method on the response to `User.findById` in the test for `findEvent/:id`. Just two little issues:

1. I think the way you're describing the behavior of `/api/checkins/findEvent/:id` is incorrect. Based on my reading of the code (`checkins.router.js` lines 30-24), this endpoint finds all of the checkins whose `eventId` field match the id passed in the url. Then they populate those documents with with the corresponding users by the `userId` field. So what this endpoint should really return is "a list of users who have checked in to a specified event." I went hunting through the client to see where this endpoint is used, and it's only ever called from `ProjectLeaderDashboard.jsx`, which you worked on in issue #1713 . At that time @trillium said that this component is unused. So all in all, the stakes are low for having a super good test on this endpoint, but I think we should correct the description of what it returns and there seem to be some unneeded complexities to the test you wrote. I put more specific comments for those.
2. In the create test, it would be ideal if we could make sure `Checkin.create` is actually called with `newCheckin` and not that it's just called at all. I know that my PR does the same thing. I should figure out a way to fix that. It seems to be a slightly different issue when the request is being passed to a controller, but in this case, configuring body parser middleware seems to be the solution. See my comments on the code.

I'm learning a lot as I go through this, so again, I just want to say great work.

jng34:
Hey @dannyprikaz,

Thank you for thoroughly reviewing the PR and for the detailed feedback.

You are correct in the description for the `/api/checkins/findEvent/:id` test and for properly testing to see if the info passed in the body for the `CheckIn.create` method.

All changes have been made. 

As always, I appreciate the guidance.

jng34 (commented):
Please see updated changes.

JackHaeg:
@dannyprikaz Can you please review the updates from @jng34 on this PR and potentially approve when you have a moment this week?

dannyprikaz (approved):
Looks good! thanks

PR #1913, opened 2025-05-03
Unit testing for Questions router

### LINKED ISSUE #1893: Create New Unit Tests for ./backend/routers/questions.router.js
### Overview
Create unit tests for each route on `./backend/routers/questions.router.js`

### Action Items
- [x] Create tests for every route described in `./backend/routers/questions.router.js`, isolating their functionality from all dependencies

### Resources/Instructions
Example: https://github.com/hackforla/VRMS/pull/1882/files
- This issue is part of the epic: https://github.com/hackforla/VRMS/issues/1880

jng34:
Fixes #1893 

  - Wrote unit tests for `questions.router`

### Screenshots of Passed Tests  
image


backend/routers/questions.router.test.js
```diff
@@ -0,0 +1,115 @@
...
+        answers: {
+          answerOneText: 'Pizza',
+          answerTwoText: 'Cheeseburger',
+          answerThreeText: 'Sushi',
+          answerFourText: 'Chicken',
+        },
+      },
+    ];
+
+    it('should return all questions with GET /api/questions', async (done) => {
```
dannyprikaz, line 50: Because the logic for responding to requests in present in this router, we should test for fail cases. After this test for a successful `Question.find()`, we should add something like this:

```
    it('should return 400 status code when there is an error with GET /api/questions', async (done) => {
      // Mock the Question.find() method
      Question.find.mockRejectedValue(new Error('Database error'));

      // Mock the request to the API
      const response = await request.get('/api/questions');

      // Tests
      expect(Question.find).toHaveBeenCalled();
      expect(response.status).toBe(400);

      // Marks completion of tests
      done();
    });
    ```

File: backend/routers/questions.router.test.js [hunk 2]
```diff
@@ -0,0 +1,115 @@
...
+      // Tests
+      expect(Question.find).toHaveBeenCalled();
+      expect(response.status).toBe(200);
+      expect(response.body).toEqual(mockQuestions);
+
+      // Marks completion of tests
+      done();
+    });
+
+    it('should return a specific question with GET /api/questions/:id', async (done) => {
```
dannyprikaz, line 88: Need a similar fail-case test here.

backend/routers/questions.router.test.js
```diff
@@ -0,0 +1,115 @@
...
+      expect(response.status).toBe(200);
+      expect(response.body).toEqual(mockQuestion);
+
+      // Marks completion of tests
+      done();
+    });
+  });
+
+  describe('CREATE', () => {
+    it('should create a new question with POST /api/questions/', async (done) => {
```
dannyprikaz, line 133: Need a similar fail case test here

dannyprikaz (changes requested):
Hey, James. These tests look good for the success scenarios,, but like we discussed with Angela last week, when the logic we are testing has failing scenarios, we should add tests for those as well. I included an example in my comments.

jng34 (commented):
@dannyprikaz 
Added testing for all fail cases.

dannyprikaz (approved):
Hey, @jng34 . This all looks good to me. As a note for the future, I think this code could be a little DRYer. This applies to the last PR I approved as well, for the recurring events. We could define a constant for `mockDatabaseError` or something like that up top. Then we could use that as the `mockRejectedValue` for `Question.find` and `Question.create`. That would save a couple of lines. More impactful, I think we could define `consoleLogSpy` up top and then move `consoleLogSpy.mockRestore()` to the `afterEach` block, if that `mockRestore` call isn't already covered by `jest.clearAllMocks()`

Something to think about!

PR #1933, opened 2025-05-18
enable enter key submission on login

### LINKED ISSUE #1919: Enable "Enter" Key Submission on Login Forms
### Overview
Allow users to press "Enter" to submit their email address when logging in as an admin or (potentially) as a returning user checking in to a meeting. Currently, the "Enter" key does not trigger form submission, creating a frustrating user experience.

### Description:
- On Dev build, admin goes to login here: https://dev.vrms.io/login.
- Admin user types email address to login and presses Enter. 
  - BUG: Currently, on [dev](https://dev.vrms.io/login), this results in no submission of the email address. 
  - On [PROD](https://www.vrms.io/login), pressing enter results in the submission of the email.  
- See screen recording in resources section for more info.
- Check to see if this same behavior also occurs when a returning user is checking in to the meeting (see "Action Items" below for more info) 

### Action Items
- [ ] Investigate and confirm whether the issue is due to the submit button not being nested inside a <form> element.
  - [ ] If this is the root cause of the problem, then adjust the file used for the admin login, so the submit button is correctly nested within a `<form>`, enabling "Enter" key submission. (Believe this is the file: `./src/components/auth/Auth.js` - recommend double checking)
  - [ ] Investigate if the same "Enter" submission issue is also occurring on the `returnUserForm` component on a local dev build that [includes this PR](https://github.com/hackforla/VRMS/pull/1875)
    - [ ] If the same issue is occurring on this page, then apply the same fix to ensure consistent behavior across both login flows (`./src/components/presentational/returnUserForm.js`)
- [ ] Test both login flows to confirm the "Enter" key triggers the expected submission action.

### Resources/Instructions
<details><summary>Screen Recording showing user pressing the "Enter" key after submitting on DEV and then PROD</summary>
<p>

https://github.com/user-attachments/assets/53d7a5a9-1c76-46da-895e-5ee214743652

</p>
</details> 
- Note: This bug probably has to do with the input field and the button not being in the same form block

angela-lee1:
Fixes #1919

### What changes did you make and why did you make them?

  - Enable enter key submission on the admin login page

### DIFF
--- a/client/src/components/auth/Auth.jsx
+++ b/client/src/components/auth/Auth.jsx
@@ -105,36 +105,33 @@ const Auth = () => {
             Welcome Back!
           </Typography>
         </div>
-        <FormControl
+        <form
           onSubmit={handleLogin}
           className="form-check-in"
           autoComplete="off"
         >
-          <div className="form-row">
-            <div className="form-input-text">
-              <Box className="form-row">
-                <Box className="form-input-text">
-                  <TextField
-                    label="Enter your email address:"
-                    type="email"
-                    name="email"
-                    placeholder="Email Address"
-                    required="required"
-                    onChange={handleInputChange}
-                    onKeyDown={(e) => {
-                      if (e.key === 'Enter') {
-                        handleLogin(e);
-                      }
-                    }}
-                    aria-label="Email Address"
-                    data-test="input-email"
-                    autoComplete="email"
-                  />
+          <FormControl>
+            <div className="form-row">
+              <div className="form-input-text">
+                <Box className="form-row">
+                  <Box className="form-input-text">
+                    <TextField
+                      label="Enter your email address:"
+                      type="email"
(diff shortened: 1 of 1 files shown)


trillium (changes requested):
He Angela, thanks for your work here.

This looks a little overengineered, overall we prefer to continue to use built in HMTL functionality to submit the form. It looks like we just need to wrap the FormControl component and a form component and then we should be good.

Make sure to move the properties from the FormControl component into the form component

```js
// Move from <FormControl> to <form>
          onSubmit={handleLogin}
          className="form-check-in"
          autoComplete="off"
```

angela-lee1:
@trillium,  @JackHaeg, I made changes per your comments. Please take a look at it when you have a chance. Thanks.

trillium (approved):
Thanks for making these changes, approved :)

PR #1955, opened 2025-06-04
Unit testing for User model

### LINKED ISSUE #1951: Create New Unit Tests for ./backend/models/user.model.js
### Overview
Create unit tests for behavior that we are depending on in `user.model.js`

### Action Items
- [x] Create a test for the serialization of user model
- [x] Create a test for the validation of a user model

### Resources/Instructions
https://mongoosejs.com/docs/models.html

- This issue is part of the epic: #1950

jng34:
Fixes #1951

  - Wrote unit tests for `User` model
  - Checked for serialization and validation

### Screenshot of passed tests
passed-tests


backend/models/user.model.test.js
```diff
@@ -0,0 +1,165 @@
+// import necessary modules
+const mongoose = require('mongoose');
+const { User } = require('./user.model');
+// MongoDB Memory Server is used for testing purposes to avoid using a real database
+const { MongoMemoryServer } = require('mongodb-memory-server');
+
+
+describe('Unit tests for User Model', () => {
+  let mongoServer;
```
dannyprikaz: We should remove all of this setup for the mongoServer because this was the source of intermittent errors previously.

backend/models/user.model.test.js
```diff
@@ -0,0 +1,165 @@
...
+    it('should fail validation check if accessLevel is invalid', async () => {
+      // Create a mock user with an invalid accessLevel
+      const mockuser = new User({
+        accessLevel: 'projectleader', // not 'user', 'admin', or 'superadmin'
+      });
+
+      // Attempt to validate the mock user by checking for valid accessLevel
+      let error;
+      try {
+        await mockuser.save();
```
dannyprikaz: This should be changed to `await mockuser.validate();` so as to not require a mongo server

backend/models/user.model.test.js
```diff
@@ -0,0 +1,165 @@
...
+          lastName: 'User',
+        },
+        email: 'mockuser@gmail.com',
+        accessLevel: 'user',
+      });
+
+      // Attempt to save the mock user
+      let error;
+      try {
+        await mockUser.save();
```
dannyprikaz: This should be changed to `await mockUser.validate();`

backend/models/user.model.test.js
```diff
@@ -0,0 +1,165 @@
...
+        githubPublic2FA: mockUser.githubPublic2FA,
+        availability: mockUser.availability,
+        managedProjects: mockUser.managedProjects,
+        isActive: mockUser.isActive,
+      });
+    });
+  });
+
+  describe('Validation test', () => {
+    it('should fail validation check if user email is not unique', async () => {
```
dannyprikaz: This test cannot be accomplished using the `new` keyword and `.validate()` instead of `.save()`, but I've learned a few things.

https://mongoosejs.com/docs/validation.html#the-unique-option-is-not-a-validator

The unique option is not a validator. See the above link for more info.

mockingoose is a library that is used for reliable unit tests of mongoose, because it mocks the outcomes of your mongoose operations without creating a real databbase.
https://www.npmjs.com/package/mockingoose

Here's an example from Google's AI for testing a unique value using mockingoose:
```
// user.model.js
const mongoose = require('mongoose');

const userSchema = new mongoose.Schema({
  email: { type: String, unique: true, required: true },
});

const User = mongoose.model('User', userSchema);

module.exports = User;

// user.test.js
const mockingoose = require('mockingoose');
const User = require('./user.model');

describe('User Model', () => {
  beforeEach(() => {
    mockingoose.resetAll();
  });

  it('should save a user with a unique email', async () => {
    const newUser = new User({ email: 'test@example.com' });
    mockingoose.User.toReturn(newUser, 'save');
    const savedUser = await newUser.save();
    expect(savedUser.email).toBe('test@example.com');
  });

  it('should throw an error when saving a user with a duplicate email', async () => {
    const existingUser = new User({ email: 'test@example.com' });
    mockingoose.User.toReturn(existingUser, 'findOne');
    const newUser = new User({ email: 'test@example.com' });
    mockingoose.User.toReturn(Promise.reject({ code: 11000 }), 'save'); // Simulate duplicate key error
    await expect(newUser.save()).rejects.toThrowError('E11000 duplicate key error');
  });
  
  it('should throw an error when updating with a duplicate email', async () => {
    const existingUser = new User({ email: 'test@example.com' });
    mockingoose.User.toReturn(existingUser, 'findOne');
    const userToUpdate = new User({email: 'oldemail@example.com'});
    userToUpdate.email = 'test@example.com';
    mockingoose.User.toReturn(Promise.reject({ code: 11000 }), 'save');
    await expect(userToUpdate.save()).rejects.toThrowError('E11000 duplicate key error');
  });
});
```

dannyprikaz (changes requested):
Hey, James. Sorry for the late review on this. Maybe we should discuss in the meeting tonight, but I think that my comments should provide the right approach to unit testing models without creating a `MongoMemoryServer`

backend/models/user.model.test.js
dannyprikaz: The AI solution from `it('should throw an error when saving a user with a duplicate email'` didn't wind up working for me when I tried it out, but reviewing the documentation for mockingoose, I got it to work like this:

```
      const existingUser = new User({
        email: 'test@example.com'
      });
      mockingoose.User.toReturn(existingUser, 'findOne');
      const newUser = new User({
        email: 'test@example.com'
      });
      mockingoose.User.toReturn(new Error({ code: '11000'}), 'save');

      newUser.save().catch(err => {
        // Tests
        expect(err).toBeDefined();
        expect(err.code).toBe('11000'); // MongoDB duplicate key error code
      });
```

jng34: [Reference](https://www.perplexity.ai/search/what-is-the-unique-true-proper-02V8tsDMQFyp6wxTSHissw#1)

I did a bit of digging with AI and found a solution where you do not need to use an in-memory MongoDB server. Basically, I just mocked the `findOne` and `create` methods and performed a manual level validation check using an `async` function `createUser` to mimic the Mongoose validation (see updated code). 

Please let me know what you think of this method.

dannyprikaz: Hey James,

I don't think that this method actually tests anything. You are creating expectations of what will happen when you call `createUser`, the async function that's right there in your test. That doesn't actually test any part of `user.model.js`

I think that we ultimately decided that we would test for the functionality of the `unique: true` tag in the integration test because we can actually use a db there. There is a test in `backend/test/user.integration.test.js` that covers this, so I think we should go ahead and delete this test. Then this should be ready to merge.

dannyprikaz (changes requested):

backend/models/user.model.test.js
jng34: Done.

jng34 (commented):
Implemented the requested changes.

JackHaeg:
@dannyprikaz Can you please review this PR when you have a moment? Once reviewed / merged, we can then unblock work on #1979

JackHaeg:
Hi @dannyprikaz! Just checking in again - Can you please review this PR when you have a moment? Once reviewed / merged, we can then unblock work on https://github.com/hackforla/VRMS/pull/1979

dannyprikaz (approved):
Looks good!

PR #1959, opened 2025-06-10
fix: Change conditional to show managed projects list in an else block

### LINKED ISSUE #1958: Bug: Admins with PM role see filtered list instead of full project list
### Overview
Admins who are also listed as PMs are seeing a filtered projectList that only includes their PM-assigned projects, rather than the complete list they should have access to as admins. The admin-level access is being unintentionally overridden.

### Action Items
- [x] Investigate where projectList is being set and determine if role precedence is being handled correctly.
- [x] Update the logic so that if a user has both roles, admin access takes precedence over PM filtering.

### Resources/Instructions
<details><summary>Code Screenshot</summary>
<p>

Image

</p>
</details>

trillium:
Fixes #1958

- Shift managedProjects condition into and else block
  - This was overwriting projectList if the user was an `admin` and had projects they managed.

### DIFF
--- a/client/src/pages/ProjectList.jsx
+++ b/client/src/pages/ProjectList.jsx
@@ -40,12 +40,13 @@ export default function ProjectList({ auth }) {
       async function fetchAllProjects() {
         let projectData;
 
-        if(user?.accessLevel === 'admin' || user?.accessLevel === 'superadmin') {
+        if (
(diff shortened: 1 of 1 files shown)


JackHaeg (approved):
Looks good @trillium !

PR #1970, opened 2025-06-18
Upgrade vite

trillium:
This upgrade is necessary for the frontend build action


JackHaeg (approved):
Thanks for the upgrade! Approved

PR #1972, opened 2025-06-19
fix: Add arg gmail_remove_subaddress to .normalizeEmail

### LINKED ISSUE #1973: Bug: Email login fails when address contains hyphens
### Overview
Currently, email addresses with hyphens (e.g., `vrms-admin-perm@hackforla.org`) are not being recognized by the system during login. As a result, the magic login email is not being sent to the user. We need to identify and fix the logic preventing proper handling of these valid email formats.

### Action Items
- [x] Reproduce the issue by attempting login with `vrms-admin-perm@gmail.com`.
- [x] Investigate email validation logic or filtering that may incorrectly reject or ignore hyphenated emails.
- [x] Fix the logic to ensure all valid email formats, including those with hyphens, are accepted.
- [x] Test login with multiple email formats to confirm the fix. (note - we will not be able to test until fix is deployed to Dev)
- [ ] If applicable, add a unit or integration test for hyphenated email login.

### Resources/Instructions
- Check backend email validation logic (`authController`, `userService`, or equivalent).
- Verify whether frontend form validation may also be blocking hyphenated inputs.
- Reference standard email regex validation patterns: [HTML Email Validation Spec](https://html.spec.whatwg.org/multipage/input.html#e-mail-state-(type=email))

trillium:
Fixes #1973 

email error bug

### DIFF
--- a/backend/validators/user.api.validator.js
+++ b/backend/validators/user.api.validator.js
@@ -6,7 +6,7 @@ async function validateCreateUserAPICall(req, res, next) {
   await body('email', 'Invalid email')
     .exists()
     .isEmail()
-    .normalizeEmail({ gmail_remove_dots: false })
+    .normalizeEmail({ gmail_remove_dots: false, gmail_remove_subaddress: false })
     .run(req);
 
   // Finds the validation errors in this request and wraps them in an object with handy functions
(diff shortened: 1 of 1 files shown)


dannyprikaz (approved):
lgtm

PR #1982, opened 2025-07-15
userPermissionFixProjectMembers

jng34:
Fixes revert issue


trillium (approved):
Approving these changes, discussed on Monday's call

Issue #1665, opened 2024-06-11
Update HTML components to MUI: ./src/components/manageProjects/editMeetingTimes.jsx


trillium:
### Dependency: 
- [x] #1724 

### Overview
Review all components in`./src/components/editMeetingTimes.jsx` and replace all standard HTML components with applicable MUI components.

### Action Items
- [x] Review components in:`./src/components/editMeetingTimes.jsx` and search for any standard HTML components.
  - _IF all components are already updated to relevant MUI_
    - [x] Comment on this issue that all components are already updated to MUI. 
    - [ ] Close the issue & check the box associated with this issue on the [MUI issue epic](https://github.com/hackforla/VRMS/issues/1617)
    - [ ] Move on to the next [MUI issue](https://github.com/hackforla/VRMS/issues/1617) 
  - _IF standard HTML components are found in the file:_ 
    - [ ] Review the following PR https://github.com/hackforla/VRMS/pull/1702 that addressed multiple issues (including this one). This PR can serve as a starting point for this issue.
    - [ ] Replace these standard HTML components with applicable MUI components. 
    - [ ] Make a PR for this file

### Resources
- https://mui.com/material-ui/
- If you run into any questions, add a comment to the issue and tag "@trillium" 
- OLD closed PR: https://github.com/hackforla/VRMS/pull/1702 that addressed multiple issues (including this one). This PR can serve as a starting point for this issue.
   - When referencing old PR above, be sure to use `.jsx` instead of `.js` in PR
- This issue is part of the epic: #1617

JackHaeg:
**Update:** 
- Closed https://github.com/hackforla/VRMS/pull/1702 as this PR was out of date and included duplicate work for multiple issues. Moving this issue back to the prioritized backlog. 
- Recommend reviewing https://github.com/hackforla/VRMS/pull/1702 and using this as a jumping off point when submitting a fresh PR for this issue.

bconti123:
I reviewed `./src/components/manageProjects/editMeetingTimes.jsx`, and it is already updated to the MUI. I did not find any standard HTML components that need replacement.

PR #2030, opened 2025-10-20
1724 - Add MUI Modal open/close behavior

### LINKED ISSUE #1724: Change Edit Meeting Times base component to an MUI modal
### Overview
Replace the current modal in `EditMeetingTimes` with an accessible modal from [MUI components](https://mui.com/material-ui/react-modal/)

### Action Items
 - [x] Review the following PR: https://github.com/hackforla/VRMS/pull/1741. This PR was created for this issue and was closed due to merge conflicts and changes that will be difficult to overcome due to more recent edits on VRMS. This PR can serve as a starting point for this issue. 
 - [x] Replace old modal with accessible modal ([link to file](https://github.com/hackforla/VRMS/blob/development/client/src/components/manageProjects/editMeetingTimes.js#L95))
 - [x] Ensure close functionality works
 - [x] Provide screenshots of before and after
 - [ ] Once this issue is closed successfully, remove the dependency from the following issue and move to the prioritized backlog (recommend moving on to this issue) #1665

### Resources/Instructions
 - [MUI React Modal](https://mui.com/material-ui/react-modal/)
 - OLD closed PR for this issue to review: https://github.com/hackforla/VRMS/pull/1741
   - When referencing old PR above, be sure to use `.jsx` instead of `.js` in PR
 - This issue is part of the epic: #1617

bconti123:
Fixes #1724 

  - Ensured the open/close functionlity of the MUI Modal works correctly.
  - Implemented closing behavior for the Escape key, Close button, and clicking outside the modal box.
  - Created this PR to avoid merge conflicts.
  - After this PR is merged, #1665 will be unblocked. 

<details>

ModalsBefore

</details>

<details>
  
ModalsAfter

</details>

### DIFF
--- a/client/src/components/manageProjects/editMeetingTimes.jsx
+++ b/client/src/components/manageProjects/editMeetingTimes.jsx
@@ -23,6 +23,8 @@ const EditMeetingTimes = ({
   useEffect(() => {
     if (selectedEvent) {
       setOpen(true);
+    } else {
+      setOpen(false);
     }
   }, [selectedEvent]);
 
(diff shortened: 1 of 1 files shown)


bconti123:
Thank you for the review! I've made the updates and resolved the conversation.

client/src/components/manageProjects/editMeetingTimes.jsx
```diff
@@ -69,58 +81,71 @@ const EditMeetingTimes = ({
...
-        type="button"
-        className="meeting-cancel-button"
-        onClick={() => {
-          setFormErrors(null);
-          setSelectedEvent(null);
+    <Modal
+      open={open}
+      onClose={handleClose}
+      aria-labelledby="edit-meeting-modal-title"
+      aria-describedby="edit-meeting-modal-description"
```
bkmorgan3: @bconti123  I didn't see an id that ties these aria attributes to other elements.  If there is no title or description in the Modal you can remove these.

bconti123: @bkmorgan3 Thanks for pointing that out. I'll review the aria attributes and remove them if they're not needed.

bconti123: I removed the aria attributes since I didn't find any matching element IDs. Conversation resolved.

bkmorgan3 (approved):

bkmorgan3:
@bconti123 approved, good to merge

eunicode (approved):
Great work! Thanks for your patience while I reviewed this.

Question (non-blocking): Was there a reason you removed ARIA attributes?

Suggestion (non-blocking): You can simplify this component by removing both the useEffect and the `open` state. Currently, the `open` state syncs with the `selectedEvent` state, so we’re maintaining two states that represent the same piece of information. Instead of duplicating state, we can derive it. useEffect is also not necessary, because React will automatically re-render when `selectedEvent` changes. 

Current approach:
```javascript
const [open, setOpen] = useState(false);

useEffect(() => {
  if (selectedEvent) { setOpen(true); }
  else { setOpen(false); }
}, [selectedEvent]);
```

Suggested approach (derived state):
```javascript
const open = Boolean(selectedEvent);
```

`open` is now derived from `selectedEvent` and the useEffect can be removed entirely.

Your modal would stay the same:
```javascript
<Modal open={open} onClose={handleClose}>
```

To close it, you only need to reset `selectedEvent`, so handleClose() would just be:
```javascript
function handleClose() {
  setFormErrors(null);
  setSelectedEvent(null);
}
```

When `selectedEvent` becomes null, the component re-renders, `open` becomes false, and the modal closes. This avoids duplicated state and follows React best practices around deriving state instead of duplicating state.

bconti123:
@eunicode Thank you for your feedback. I like the suggested approach. This makes sense. You are right that the `open` state and the `useEffect` were duplicating the same information from `selectedEvent`, and I updated the code to look cleaner.

Regarding your question about the ARIA attributes, yes, there was a reason in the reviewed changes, and that conversation is now resolved. I did not find any IDs that matched the ARIA attributes, so I resolved the thread.

I will merge this now.

Issue #2075, opened 2026-01-27
CSRF protection for cookie-based refresh


geolunalg:
### Overview
User Story:
As a system, I want to prevent CSRF attacks on endpoints that rely on cookies.

### Action Items
Acceptance Criteria:
- [x] Refresh/logout endpoints are protected by:
    - [x] SameSite cookie settings and/or
    - [x] CSRF token (double submit or header-based)
- [x] Backend rejects refresh requests missing CSRF proof (if implemented).

### Resources/Instructions
- This issue is part of the epic: #2065

geolunalg:
@JackHaeg @trillium @rteas 

This is dependent on #2067 and #2068 client (frontend) inplementation portion

Size: L

rteas:
Closed under https://github.com/hackforla/VRMS/pull/2090

Issue #2072, opened 2026-01-27
Logout (single device)


geolunalg:
### Overview
User Story:
As a user, I want to log out so my session is ended on this device.

### Action Items
Acceptance Criteria:
- [ ] POST /auth/logout:
    - [ ] revokes the refresh token record server-side
    - [ ] clears the refresh token cookie
- [ ] Subsequent refresh attempts fail with 401.

### Resources/Instructions
- This issue is part of the epic: #2065

rteas:
Backend Task completed. Orignal /auth/logout api has been modified to remove refresh token, preventing a user from re-using the refresh token.

Frontend must remove its JWT manually.

geolunalg:
@JackHaeg @trillium @rteas 

Need confirmation that the backend is refresh token is invalidated on the back end form Richy. 

Frontend: On logout shoudl return to the login page

Size: S to M

rteas:
Closed under https://github.com/hackforla/VRMS/pull/2090

Issue #2068, opened 2026-01-27
Refresh access token (silent)


geolunalg:
### Overview
User Story:
As a user, I want my session to continue seamlessly when my access token expires.

### Action Items
User Story:
As a user, I want my session to continue seamlessly when my access token expires.
Acceptance Criteria:
- [ ] When access token is expired and API returns 401, client calls POST /auth/refresh.
- [x] Backend reads refresh token from cookie and:
    - [x] if valid: returns a new access token
    - [x] if invalid/expired: returns 401 and client routes to login
- [ ] Client retries the original request once after successful refresh.

### Resources/Instructions
- This issue is part of the epic: #2065

rteas:
Backend api has JWT refresh functionality.

For a seamless UX, we should have the frontend refresh the token roughly a minute before the token expires.
We can probably implement a timer function to check every ~30s.

geolunalg:
@JackHaeg @trillium @rteas 

Need confirmation from Richy if this is done `When access token is expired and API returns 401, client calls POST /auth/refresh.`

Frontend task `Client retries the original request once after successful refresh.`

Size: M

rteas:
closed under https://github.com/hackforla/VRMS/pull/2090

Issue #2066, opened 2026-01-27
Login


geolunalg:
### Overview
User Story:
As a user, I want to sign in with my credentials so I can access protected features.

Acceptance Criteria:
- [x] Given valid credentials, when I log in, then:
    - [x] backend returns an access token
    - [x] backend sets a refresh token cookie (HttpOnly, Secure, SameSite)
    - [x] response includes user profile basics or client can call /me
- [x] Given invalid credentials, when I log in, then I receive a generic error (no user enumeration).

### Resources/Instructions
- This issue is part of the epic: #2065

rteas:
Backend /auth/verify-sign has been modified to create refresh token and also create access token (JWT) for client containing relevant user info (roles, email, etc).

Will return a 401 for front end to handle when invalid credentials are provided.

rteas:
Completed

PR #2090, opened 2026-02-24
Refactor permission management

### LINKED ISSUE #1861: Refactor Permission Management to a Middleware-Based Solution
### Overview
The current VRMS permission system (`accessLevel`) relies on hardcoded values / string matching checks (e.g., `is_admin`), which limits flexibility and maintainability. To enhance scalability and streamline permission management, permissions should be refactored into a middleware-based solution. This will allow for centralized handling and easier updates across the application.

### Action Items
- [ ] Research approaches for refactoring the current permission system into a middleware-based solution.
  - Ensure that the new implementation supports upcoming User Permission Search features, which are set to be merged in [#1737](https://github.com/hackforla/VRMS/issues/1737).
- [ ] Design and document middleware architecture, including:
  - Permission validation workflows.
  - Integration points with existing authentication mechanisms.
- [ ] Create diagrams outlining the proposed middleware solution to illustrate system interactions and permission handling.

### Resources/Instructions
- Review the related issue that uncovered this problem: #1802.
- Ensure alignment with existing authentication and authorization best practices.

rteas:
Fixes 
#1861 
#2066 
#2068 
#2072 
#2075 

  - Refactored middleware/backend to use refresh tokens and json web tokens (jwts) - see auth.middleware.js, user.routes.js
  - Refactored frontend code to utilize most frontend code jwt via authContext.jsx
  - Both middle/back/front rely on the single source of truth (auth) in the shared folder with the following files: roles.js and authorizationUtils.js

<details>
<summary>No visuals, but website runs using access tokens (JWTS) 15m usage and Refresh Tokens (allowing users extended sessions up to 30 days without needing to relog into the application). Logging out will delete the refresh token and require users to re-login.</summary>


JackHaeg:
Note - once this PR is merged, hold off on rebuilding DEV until check-in for logged in users is enabled [add issue link here]. Otherwise, we may run into an issue where most users remain logged in and cannot check in without first logging back out of VRMS

JackHaeg:
@rteas mentioned that Project manager permissions are still handled the same way as before (standard user who is associated with a project = project manager). @rteas may create a new issue that creates a new UAL of "Project Manager" to follow the middleware refactor for other UALs (admin, superadmin, users, and project managers). Requires front end changes. Also, users assigned to an existing project will need to be promoted to "project manager" UAL in database.

backend/config/auth.config.js
```diff
@@ -1,8 +1,10 @@
-/*eslint-disable */
 module.exports = {
   SECRET:
     'c0d7d0716e4cecffe9dcc77ff90476d98f5aace08ea40f5516bd982b06401021191f0f24cd6759f7d8ca41b64f68d0b3ad19417453bddfd1dbe8fcb197245079',
   CUSTOM_REQUEST_HEADER: process.env.CUSTOM_REQUEST_HEADER,
-  TOKEN_EXPIRATION_SEC: 900,
+  // 15 minutes
```
geolunalg: what is the difference between these two ACCESS_TOKEN_EXPIRATION, and ACCESS_TOKEN_EXPIRATION_MS?

backend/middleware/auth.middleware.js
```diff
@@ -1,28 +1,156 @@
...
+const { RefreshToken, User } = require('../models');
+const crypto = require('crypto');
+const AuthUtils = require('../../shared/authorizationUtils');
+
+const SECRET_KEY = process.env.JWT_SECRET;
+
+// Utility functions
+
+function generateAccessToken(user, auth_origin) {
+  return jwt.sign(
```
geolunalg, line 13: Why is there no experation time on the signature of the access token?

backend/app.js
```diff
@@ -1,11 +1,10 @@
- 
 // app.js - Entry point for our application
 
 // Load in all of our node modules. Their uses are explained below as they are called.
 const express = require('express');
+// const cron = require('node-cron');
```
geolunalg: Why are these commented out?

backend/app.js
```diff
@@ -53,16 +52,18 @@ app.use(cookieParser());
 app.use(morgan('dev'));
 
 // WORKERS
-const runOpenCheckinWorker = require('./workers/openCheckins');
-runOpenCheckinWorker(cron, fetch);
+// const runOpenCheckinWorker = require('./workers/openCheckins')(cron, fetch);
```
geolunalg: Why are these commected out?

backend/config/auth.config.js
```diff
@@ -1,8 +1,10 @@
-/*eslint-disable */
 module.exports = {
   SECRET:
```
geolunalg: This should not be part of the code, this should be as the name implies a secret. This should be a env var.

backend/config/auth.config.js
```diff
@@ -1,8 +1,10 @@
...
     'c0d7d0716e4cecffe9dcc77ff90476d98f5aace08ea40f5516bd982b06401021191f0f24cd6759f7d8ca41b64f68d0b3ad19417453bddfd1dbe8fcb197245079',
   CUSTOM_REQUEST_HEADER: process.env.CUSTOM_REQUEST_HEADER,
```
geolunalg, line 3: Calling an env variable like this, if the var does not exist will cause a silent failure returning undefined.

backend/middleware/auth.middleware.js
```diff
@@ -1,28 +1,156 @@
...
-  // eslint-disable-next-line dot-notation
-  let token = req.headers['x-access-token'] || req.headers['authorization'];
-  if (token.startsWith('Bearer ')) {
-    // Remove Bearer from string
-    token = token.slice(7, token.length);
+const { RefreshToken, User } = require('../models');
+const crypto = require('crypto');
+const AuthUtils = require('../../shared/authorizationUtils');
+
+const SECRET_KEY = process.env.JWT_SECRET;
```
geolunalg: We should avoid calling these blindly, they will silently fail if they do not exist.

backend/test/old-tests/projects.router.test.js
```diff
@@ -30,25 +28,18 @@ describe('CREATE', () => {
...
     const auth_origin = 'TEST';
-    token = jwt.sign(
-      { id: user.id, role: user.accessLevel, auth_origin },
-      CONFIG_AUTH.SECRET,
-      {
-        expiresIn: `${CONFIG_AUTH.TOKEN_EXPIRATION_SEC}s`,
-      },
-    );
-  })
+    token = jwt.sign({ id: user.id, role: user.accessLevel, auth_origin }, CONFIG_AUTH.SECRET, {
```
geolunalg: On these test why are you not using the function in code to sign the token for the test?

backend/test/old-tests/projects.router.test.js
```diff
@@ -106,14 +97,10 @@ describe('UPDATE', () => {
...
```
geolunalg: Same comment here, why is the funtion in code to sign the token for tests?

backend/test/old-tests/projects.router.test.js
```diff
@@ -194,14 +181,10 @@ describe('DELETE', () => {
...
```
geolunalg: What is token for?

backend/config/auth.config.js
rteas: I agree, I originally put this so it would work without needing env variables (and updating it later), but thinking about it now it wasn't a good idea and I should have be something more like secret = process.env.secret || 'value'. Going to fix it

backend/config/auth.config.js
rteas: I can remove ACCESS_TOKEN_EXPIRATION_MS, it was used for creating a custom expiration using ISO ms for date. But JWT has a built in expiration value that uses a different datatype (string).

backend/middleware/auth.middleware.js
rteas: Agreed, it's better to use it with || 'fallback value'

backend/middleware/auth.middleware.js
rteas: That's because I use the expiration time created from the built-in JWT expiration property, which is provided in the options. This is sent as a JSON response because we are using httpCookies, the frontend/react app cannot parse the JWT. Therefore, we send the hidden JWT and user/expiration data. Using the built-in JWT expiration our code is cleaner since we don't need to check for 'date' ourselves, the jwt.verify method will handle it for us (which we eventually call anyways)

geolunalg: I am not sure if this is correct. According to the documentaion for jsonwebtoken these is no default for expiresIn, notBefore, audience, subject, issuer. 

https://www.npmjs.com/package/jsonwebtoken#:~:text=There%20are%20no%20default%20values%20for%20expiresIn%2C%20notBefore%2C%20audience%2C%20subject%2C%20issuer.%20These%20claims%20can%20also%20be%20provided%20in%20the%20payload%20directly%20with%20exp%2C%20nbf%2C%20aud%2C%20sub%20and%20iss%20respectively%2C%20but%20you%20can%27t%20include%20in%20both%20places.

rteas: It's the alternative option on the from the doc you linked:

This means that the exp field should contain the number of seconds since the epoch.

Signing a token with 1 hour of expiration:

jwt.sign({
  exp: Math.floor(Date.now() / 1000) + (60 * 60),
  data: 'foobar'
}, 'secret');
Another way to generate a token like this with this library is:

jwt.sign({
  data: 'foobar'
}, 'secret', { expiresIn: 60 * 60 });

//or even better:

jwt.sign({
  data: 'foobar'
}, 'secret', { expiresIn: '1h' });

backend/app.js
rteas: It wasn't being used, so the build rejected it

backend/middleware/auth.middleware.js
geolunalg: Oh I see 👀. That is interesting, did not not that was also an option.

backend/app.js
geolunalg: If thery are not needed, they should be removed not commented out.

backend/app.js
geolunalg: Same here if we don't need these they should be removed.

shared/authorizationUtils.js
```diff
@@ -0,0 +1,54 @@
+import { ROLES, ROLE_HIERARCHY } from "./roles.js";
```
trillium, line 1: **[Claude]** 🚨 **Critical: Mixed module systems will crash Node.js**

This file uses ES module `import` at the top but also has a CommonJS `module.exports` block. Node.js will fail on the `import` statement unless the file is `.mjs` or `"type": "module"` is set in package.json. The backend's `require('../../shared/authorizationUtils')` call will crash at startup.

Same issue exists in `shared/roles.js`.

Suggested fix: Either use pure CommonJS (`const { ROLES, ROLE_HIERARCHY } = require('./roles.js')` + `module.exports = { ... }`) or configure the project for ESM properly.

shared/roles.js
```diff
@@ -0,0 +1,29 @@
+// constants/roles.js
+
+const ROLES = Object.freeze({
```
trillium, line 3: **[Claude]** 🚨 **Critical: Role name mismatch with database**

`SUPER_ADMIN` is defined as `"super_admin"` but the User model enum (in `user.model.js`) has `"superadmin"` (no underscore). Every `hasRole(user, ROLES.SUPER_ADMIN)` check will silently return `false` — no superadmin will ever be recognized.

Also, `PROJECT_MANAGER: "project_manager"` doesn't exist in the User model enum at all (`['user', 'admin', 'superadmin']`).

This needs to match the existing database values, or include a migration to update them.

backend/controllers/user.controller.js
```diff
@@ -267,21 +262,57 @@ UserController.verifySignIn = async function (req, res) {
   try {
     const payload = jwt.verify(token, CONFIG_AUTH.SECRET);
     const user = await User.findById(payload.id);
-    res.cookie('token', token, { httpOnly: true });
```
trillium, line 269: **[Claude]** 🚨 **Critical: `accessToken.exp` is `undefined` — `expiresAt` will be `NaN`**

`generateAccessToken()` returns `jwt.sign(...)` which is a **string**, not a decoded JWT object. `accessToken.exp` is `undefined`, so `expiresAt` will be `NaN`.

This breaks the client-side auto-refresh timer in `authContext.jsx`.

Fix: decode the token first:
```js
const decoded = jwt.decode(accessToken);
return res.send({
  user: user,
  expiresAt: decoded.exp * 1000,
});
```
(Note: `refreshAccessToken` already does this correctly!)

backend/controllers/user.controller.js
```diff
@@ -4,7 +4,13 @@ const { ObjectId } = require('mongodb');
...
-const { User, Project } = require('../models');
+const { User, Project, RefreshToken } = require('../models');
+const {
+  generateRefreshToken,
+  getClientIp,
+  hashToken,
+  generateAccessToken,
+} = require('../middleware/auth.middleware');
 
 const expectedHeader = process.env.CUSTOM_REQUEST_HEADER;
```
trillium, line 16: **[Claude]** 🚨 **Critical: `CONFIG_AUTH.SECRET` no longer exists**

`auth.config.js` renamed `SECRET` to `TOKEN_SECRET`, but this line still references `CONFIG_AUTH.SECRET`. This will be `undefined`, causing `jwt.verify()` to throw and **all magic link verifications to fail**.

Should be:
```js
const payload = jwt.verify(token, CONFIG_AUTH.TOKEN_SECRET);
```

backend/routers/projects.router.js
```diff
@@ -2,24 +2,27 @@ const express = require('express');
...
 
 const { ProjectController } = require('../controllers');
-const { AuthUtil } = require('../middleware');
+// const { Auth } = require('../middleware');
+// const { ROLES } = require('../../shared/roles');
 
+// Require user to be project manager or higher (commented out for now for current app to work succesfully without auth, will re-enable when auth is ready)
+// router.use(Auth.authUser, Auth.requireMinimumRole(ROLES.PROJECT_MANAGER));
 // The base is /api/projects
 router.get('/', ProjectController.project_list);
```
trillium, line 12: **[Claude]** ⚠️ **All auth middleware removed from project routes**

All `verifyCookie` middleware has been removed from create, update, patch, and delete routes. The comment says "will re-enable when auth is ready" but as-is, **anyone can create, modify, or delete projects without authentication**.

Even if this is intentional for development, it's risky to merge to main. Consider keeping at minimum `Auth.authUser` on mutation routes, or guard this behind an environment check.

backend/controllers/user.controller.js
```diff
@@ -193,17 +199,6 @@ UserController.delete = async function (req, res) {
...
-  return jwt.sign(
-    { id: user.id, role: user.accessLevel, auth_origin: auth_origin },
-    CONFIG_AUTH.SECRET,
-    {
-      expiresIn: `${CONFIG_AUTH.TOKEN_EXPIRATION_SEC}s`,
-    },
-  );
-}
-
 UserController.createUser = function (req, res) {
```
trillium, line 202: **[Claude]** ⚠️ **Refresh token cookie missing security flags**

The `refresh_token` cookie is set with only `httpOnly: true`. In production, it should also have:
- `secure: true` — only sent over HTTPS
- `sameSite: 'strict'` (or `'lax'`) — CSRF protection
- `path: '/api/auth'` — limits which requests send the cookie

```js
res.cookie('refresh_token', refreshToken, {
  httpOnly: true,
  secure: process.env.NODE_ENV === 'production',
  sameSite: 'strict',
  path: '/api/auth',
});
```

Same applies to the `token` cookie on the line above.

backend/models/refreshToken.model.js
```diff
@@ -0,0 +1,42 @@
...
+    id: this._id,
+    createdAt: this.createdAt,
+    expiresAt: this.expiresAt,
+    deviceInfo: {
+      ipAddress: this.ipAddress,
+      deviceType: this.deviceType,
+    },
+  };
+};
+
```
trillium, line 37: **[Claude]** ⚠️ **TTL index references wrong field name**

The index uses `expires_at` (snake_case) but the schema field is `expiresAt` (camelCase). MongoDB's TTL auto-deletion won't work — expired tokens will accumulate forever.

```js
// Should be:
refreshTokenSchema.index({ expiresAt: 1 }, { expiresAfterSeconds: 0 });
```

client/src/components/user-admin/EditUsers.jsx
```diff
@@ -16,6 +16,8 @@ import {
   InputLabel,
 } from '@mui/material';
 import '../../sass/UserAdmin.scss';
+import useAuth from '../../hooks/useAuth';
+import { ROLES } from '../../../../shared/roles';
```
trillium, line 20: **[Claude]** ⚠️ **Checks logged-in user's role instead of the user being edited**

`isAdmin()` and `isSuperAdmin()` from `useAuth()` check the **currently logged-in user**, not `userToEdit`. Previously this was:
```js
const [isAdmin, setIsAdmin] = useState(userToEdit.accessLevel === 'admin');
const isSuperAdmin = userToEdit.accessLevel === 'superadmin';
```

Now the admin toggle and disabled states are all based on the wrong user. For example, if *you* are an admin, the admin switch will show as checked regardless of which user you're editing.

This should continue checking `userToEdit.accessLevel`.

backend/workers/tokenCleanup.js
```diff
@@ -0,0 +1,17 @@
...
+      expiresAt: { $lt: new Date() },
+    });
+    console.log(`Cleaned up ${result.deletedCount} expired refresh tokens`);
+  } catch (err) {
+    console.error('Token cleanup error:', err);
+  }
+}
+
+// Run daily
+setInterval(cleanupExpiredTokens, 24 * 60 * 60 * 1000);
```
trillium, line 15: **[Claude]** ⚠️ **`setInterval` runs at module import time with no cleanup**

This starts a 24-hour interval the moment the module is `require()`d — including during tests. There's no way to stop it, which will cause test hangs and leaked timer warnings.

Consider exporting a start/stop function, or using the cron system that's already in the project:
```js
let intervalId;
function startCleanupSchedule() {
  intervalId = setInterval(cleanupExpiredTokens, 24 * 60 * 60 * 1000);
}
function stopCleanupSchedule() {
  clearInterval(intervalId);
}
module.exports = { cleanupExpiredTokens, startCleanupSchedule, stopCleanupSchedule };
```

backend/controllers/user.controller.js
```diff
@@ -267,21 +262,57 @@ UserController.verifySignIn = async function (req, res) {
...
+    const accessToken = generateAccessToken(user, payload.auth_origin);
+    const ipAddress = getClientIp(req);
+
+    await RefreshToken.create({
+      userId: user._id,
+      hash: hashToken(refreshToken),
+      deviceInfo: {
+        deviceType: req.headers['user-agent'],
+        ipAddress: ipAddress,
+      },
```
trillium, line 274: **[Claude]** 🐛 **Logout clears `token` cookie but not `refresh_token` cookie**

The refresh token document is deleted from the database, but the `refresh_token` cookie is never cleared from the browser. It will persist and continue being sent on requests.

```js
return res
  .clearCookie('token')
  .clearCookie('refresh_token')
  .status(200)
  .send('Successfully logged out.');
```

backend/test/old-tests/projects.router.test.js
```diff
@@ -18,9 +17,8 @@ headers.authorization = 'Bearer sometoken';
 
 let token;
```
trillium, line 18: **[Claude]** 🐛 **`ACCESS_TOKEN_EXPIRATION_SEC` doesn't exist in config**

The config has `ACCESS_TOKEN_EXPIRATION` (a string like `'15m'`), not `ACCESS_TOKEN_EXPIRATION_SEC`. This will produce tokens with `"undefineds"` as the expiration, which `jwt.sign` may reject or produce unexpected behavior.

Should be:
```js
{ expiresIn: CONFIG_AUTH.ACCESS_TOKEN_EXPIRATION }
```

client/src/components/auth/HandleAuth.jsx
```diff
@@ -19,43 +18,28 @@ const HandleAuth = (props) => {
     const api_token = params.get('token');
```
trillium, line 19: **[Claude]** 🐛 **Missing dependency + premature `isLoaded` on initial render**

1. `refreshAuth` is missing from the dependency array — React will warn about this.
2. When `isMagicLinkValid` is `null` (initial state before the first `useEffect` resolves), the `else` branch fires and sets `isLoaded = true`, which could briefly flash the "link is not valid" error before the actual validation completes.

Consider guarding against the initial `null` state:
```js
useEffect(() => {
  if (isMagicLinkValid === null) return; // still loading
  if (isMagicLinkValid) {
    refreshAuth();
  } else {
    setIsLoaded(true);
  }
}, [isMagicLinkValid, refreshAuth]);
```

backend/models/refreshToken.model.js
```diff
@@ -0,0 +1,42 @@
...
+    ipAddress: String,
+    deviceType: String,
+  },
+});
+
+refreshTokenSchema.methods.serialize = function () {
+  return {
+    id: this._id,
+    createdAt: this.createdAt,
+    expiresAt: this.expiresAt,
```
trillium, line 30: **[Claude]** 💡 **Minor: `serialize()` references wrong paths**

`this.ipAddress` and `this.deviceType` should be `this.deviceInfo.ipAddress` and `this.deviceInfo.deviceType` — the fields are nested under `deviceInfo` in the schema.

trillium (commented):
Review from Claude — flagging potential issues across this PR. Great work on centralizing authorization logic and adding refresh tokens! A few things to address before merging.

trillium (approved):
Resolved merge conflicts from development. The permission refactoring pattern (useAuth hooks + ROLES constants) is preserved throughout.

Changes reviewed:
- Auth checks use `isAdmin()`, `isSuperAdmin()`, `hasAnyRole()`, `hasMinimumRole()` consistently
- Development's active cron workers, async/await server init, and new UI features (EditProjectMembers, TitledBoxIFrame) are preserved
- Removed duplicate local `generateAccessToken` in user.controller.js (uses imported version from auth.middleware)
- Removed unused styled components in ProjectForm.jsx
- yarn.lock accepted from development (may need `yarn install` to regenerate)

trillium (dismissed):
Revoking previous approval — I resolved the merge conflicts but shouldn't self-approve.

Known issue to fix before re-review:
- `UserController.createUser` in backend/controllers/user.controller.js uses `await` but is declared as a sync function (missing `async` keyword)
- yarn.lock should be regenerated with `yarn install`
- Build and tests should be verified

geolunalg (approved):

pillarijake:
Code is live in development instance, need to review from QA perspective to ensure application functions correctly.

PR #2111, opened 2026-03-31
fix: rename TOKEN_SECRET to JWT_SECRET across backend, small fixes

rteas:
Fixes #2090 

  - Fixed issue where code was depending on SECRET instead of JWT_SECRET
  - async data call in user controller
  - add support for new method call in projects route

<details>

image

</details>

<details>
  
image

</details>

### DIFF
--- a/backend/routers/projects.router.js
+++ b/backend/routers/projects.router.js
@@ -9,7 +9,7 @@ const { ProjectController } = require('../controllers');
 // router.use(Auth.authUser, Auth.requireMinimumRole(ROLES.PROJECT_MANAGER));
 // The base is /api/projects
 
-import { AuthUtil } from '../middleware/auth.middleware';
+const { AuthUtil } = require('../middleware/auth.middleware');
 
 router.get('/', ProjectController.project_list);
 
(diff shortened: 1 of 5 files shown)


backend/routers/projects.router.js
```diff
@@ -8,6 +8,9 @@ const { ProjectController } = require('../controllers');
 // Require user to be project manager or higher (commented out for now for current app to work succesfully without auth, will re-enable when auth is ready)
 // router.use(Auth.authUser, Auth.requireMinimumRole(ROLES.PROJECT_MANAGER));
 // The base is /api/projects
+
+import { AuthUtil } from '../middleware/auth.middleware';
```
bconti123: I just saw an error. `Cannot find module...`. 
Can you change this to this? 

`const AuthUtil = require('../middleware/auth.middleware');`

It will work.

bconti123 (approved):

backend/routers/projects.router.js
rteas: Corrected, thanks

Issue #2036, opened 2025-10-21
Skip Failing Tests and Suites


geolunalg:
### Overview
The following test are failing and need to be skiped.

- test: ../backend/routers/users.router.test.js >> should get a specific User by UserId with GET to /api/users/:UserId through UserController
- suites: ../backend/test/user.integration.test.js >> 
  - CREATE
  - READ
  - UPDATE
  - DELETE 

### Action Items
An inventory of the test need to be conducted to determine the reason for the failures.

PR #2158, opened 2026-05-12
fix(backend): complete ESM migration — runtime crash fix + test suite conversion

trillium:
Fixes #

### What changes did you make and why did you make them?

**By @eunicode (original branch author):**

- Fixed the runtime crash caused by an incomplete ESM migration — `backend/workers/lib/generateEventData.js`, `refreshToken.model.js`, `tokenCleanup.js`, and `grantpermission.router.js` were left behind by PR #2134 still using `module.exports`/`require` while their callers had been converted to ESM `import`
- Fixed `auth.middleware.js` — switched to named imports from `authorizationUtils.js` and corrected a `hasAnyRole` spread bug where roles were being passed as an array instead of spread arguments, silently breaking role checks
- Added `backend/env.bootstrap.js` to fix a dotenv timing bug introduced by ESM — in ESM, static imports are hoisted and execute before module body code, so `dotenv.config()` in `app.js` was running _after_ config modules had already read `process.env`; this side-effect module is imported first in `server.js` to guarantee env vars load before any config module runs

**By @trillium (follow-up to complete the branch):**

- Completed the test suite migration from Jest to Vitest — PR #2134 switched the test runner but left all 16 active test files using Jest globals (`jest.mock`, `jest.fn`, `jest.spyOn`); converted all of them to Vitest equivalents with correct ESM mock path resolution, `vi.hoisted()` for TDZ safety, and `vi.importActual()` where needed
- Fixed stale and incorrect mock assertions uncovered during migration (e.g. `verifyUser.isAdminByEmail` asserted on a route that does not use that middleware; `verifyCookie` asserted on routes where it was later commented out)
- Stubbed broken legacy tests in `test/old-tests/` — these had broken import paths and required a live database; replaced with `describe.skip` stubs pointing to issue #2036

### Screenshots of Proposed Changes Of The Website (if any)

N/A — backend-only changes.


pillarijake (approved):
approving test suite changes

PR #2161, opened 2026-05-19
fix(ci): update PR workflow for Jest-to-Vitest migration

trillium:
> **Note:** This fix was generated with AI assistance (Claude Code) by @trillium.

## Problem

The CI workflow (`.github/workflows/all-PRs.yaml`) still uses Jest CLI flags and path conventions, causing all three test jobs to fail on every PR since the Vitest migration (#2134).

### Failures

1. **Backend unit tests** — `--testPathIgnorePatterns=routers` is a Jest flag. Vitest rejects it with `Unknown option`.
2. **Backend integration tests** — `vitest run backend/routers/` uses the wrong path. Inside the Docker container the cwd is `/srv/backend`, so `backend/routers/` matches nothing.
3. **Client tests** — `USER node` in `Dockerfile.client` prevents Vite from writing its temp config file (`vite.config.mjs.timestamp-*.mjs`) in the mounted volume, causing `EACCES`.

### Fixes

1. New `test:unit` script: `vitest run --exclude 'routers/**' --exclude 'test/**'`
2. New `test:integration` script: `vitest run routers/ test/`
3. Removed `USER node` + `--chown=node:node` from `client-development` Dockerfile stage (development containers don't need user restrictions, and they conflict with volume mounts)

### Verification

All tests pass locally:
- `yarn test:unit` — 46 passed, 1 skipped
- `yarn test:integration` — 57 passed, 33 skipped
- `yarn test` (full suite) — 103 passed, 34 skipped

### DIFF
--- a/.github/workflows/all-PRs.yaml
+++ b/.github/workflows/all-PRs.yaml
@@ -19,7 +19,7 @@ jobs:
       - name: Build the backend container
         run: docker compose build backend
       - name: Run backend unit test suite
-        run: docker compose run --rm backend yarn run test --testPathIgnorePatterns=routers
+        run: docker compose run --rm backend yarn run test:unit
 
   test-backend-integration:
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


kkchu791 (approved):
LGTM

PR #2180, opened 2026-06-18
fix(backend): rewrite eventTime.js with Temporal polyfill for timezone-correct operations

trillium:
## Summary

Fixes timezone and DST bugs in checkin workers by replacing ad-hoc date math with the `@js-temporal/polyfill` library.

## Changes
- Rewrites `eventTime.js` using the Temporal API for accurate LA-timezone operations
- Updates `createRecurringEvents`, `openCheckins`, and `closeCheckins` workers to use the new functions
- Extracts event generation logic into an importable module for testability
- Adds 12 new Vitest timezone test cases covering `getTodayDayLA`, `checkIfSameDayLA`, boundary conditions, and fall-back DST
- Fixes slackbot timezone handling and adds integration tests

## Testing
- [ ] All 12 new timezone tests pass
- [ ] Backend unit tests pass
- [ ] Backend integration tests pass


backend/yarn.lock
trillium, line 1: This project shouldnt have bun.lock, remove

backend/bun.lock
trillium, line 1: Remove - we do not use bun here

backend/workers/lib/recurringEventOps.js
```diff
@@ -0,0 +1,102 @@
...
+      city: loc.city ? loc.city : 'REMOTE',
+      state: loc.state ? loc.state : 'REMOTE',
+      country: loc.country ? loc.country : 'REMOTE',
+    };
+  }
+
+  return eventToCreate;
+}
+
+/**
```
trillium, line 117: This is a bad way of going about this issue, we expect that a event name change will not result in duplicate event generation, but based on the Implementation that only compares the names what is likely to happen is that if the name is changed a duplicate event will be created

backend/workers/closeCheckins.js
```diff
@@ -40,21 +40,16 @@ export default (cron, fetch) => {
       return null;
     }
   }
+
   async function sortAndFilterEvents() {
     const events = await fetchEvents();
-
-    const now = Date.now();
+    const { isPastCloseWindow } = await import('./lib/eventTime.js');
```
trillium: This is a bad way of dealing with imports, please import at the top or split into separate files

backend/workers/closeCheckins.js
```diff
@@ -40,21 +40,16 @@ export default (cron, fetch) => {
...
 
     if (events && events.length > 0) {
```
trillium, line 46: Thjis comparison can probably be simpler and more readable

backend/workers/lib/eventTime.js
```diff
@@ -0,0 +1,149 @@
...
+}
+
+/**
+ * Check if 3 hours have passed since the event started.
+ * Uses UTC instants for comparison — no timezone conversion needed.
+ *
+ * @param {Date|string} eventDate - Event start time
+ * @param {Date|string} now - Current time
+ * @returns {boolean}
+ */
```
trillium, line 164: This function is a bit hardcoded, we are open to making the params such as 3 hours past clsoe time something that sits in a config instead

backend/workers/openCheckins.js
trillium, line 53: same comparison, confusing as written , can be simpliefied

backend/workers/timezone.test.js
```diff
@@ -0,0 +1,278 @@
...
+    // Both are UTC Tuesday Jan 7, but different LA calendar days
+    const lateLAMonday   = new Date('2025-01-07T07:58:00Z'); // 11:58pm PST Mon Jan 6
+    const earlyLATuesday = new Date('2025-01-07T08:02:00Z'); // 12:02am PST Tue Jan 7
+
+    expect(checkIfSameDayLA(lateLAMonday, earlyLATuesday)).toBe(false);
+  });
+});
+
+describe('isInOpenWindow — exact boundary behavior', () => {
+  test('event exactly at now is in the open window', () => {
```
trillium, line 220: can we add comments here for human readable explanation of each timestamp and why we test these timestamps

kkchu791 (approved):
LGTM

